PR #3, opened 2019-03-28
Feature dropdown component

ben-schumacher:


spiegela:
This is looking close, but I noticed a couple of things:

1.  I noticed that if I expand sub-menus, and then close the top button, the sub menus are left open.
Dropdown-close-cascade

2. I checked in the library app, and it appears to support "clicking away" to close the component. I'm not sure how to implement that in React, but we should take a look.

3. The library app isn't clear, but I think the button icon should toggle between up and down when the dropdown is expanded. 

The last piece is more of an aside. I see that these drop downs don't look like they really support scrollable containers.  Since things in Relay are all paginated, we should consider adding scrolling and search support to this.

ben-schumacher:
Hi @spiegela,
Items 1 and 2 above are actually working for me as intended.  I can't reproduce item 1, and for item 2, it's always been my intention that the dropdown goes back to it's original state when clicking away.  For item 3, I just pushed a commit that switches the caret up/down when clicking in/out of the dropdown.

Ben

PR #11, opened 2019-05-14
Feature datepicker component

ben-schumacher:


src/forms/datepicker/DatePicker.tsx
```diff
@@ -0,0 +1,423 @@
...
+
+enum ViewMode {
+    Day,
+    Month,
+    Year,
+}
+
+export class DatePicker extends React.PureComponent<DatePickerProps, DatePickerState> {
+    static defaultProps: DatePickerProps = {
+        locale: "en",
```
alexander-rassanov: you forgot please other props here that are not required

src/forms/datepicker/DatePicker.tsx
```diff
@@ -0,0 +1,423 @@
...
+                              Moment.localeData(this.props.locale).longDateFormat("L"),
+                          )
+                        : this.state.navValue,
+            },
+            this.afterToggle,
+        );
+    }
+
+    afterToggle = () => {
+        if (this.state.isOpen) this.subscribeDocumentClick();
```
alexander-rassanov: Let's be more consistent. Add brackes
if () {
} else {
}

src/forms/datepicker/DatePicker.tsx
```diff
@@ -0,0 +1,423 @@
...
+                                                        return (
+                                                            <td key={"day_" + day} className="calendar-cell weekday">
+                                                                {day.substring(0, 1)}
+                                                            </td>
+                                                        );
+                                                    })}
+                                                </tr>
+                                            </tbody>
+                                        </table>
+                                        <table className="calendar-table calendar-dates">
```
alexander-rassanov: Table adds nesting. Can we move it to other components to make render more readable and encapsulated?

src/forms/datepicker/DatePicker.tsx
```diff
@@ -0,0 +1,423 @@
...
+                                                                const isSelected = calendar.isSame(moment, "day"); // are we on selected?
+                                                                const isDisabled = !navValue.isSame(calendar, "month");
+                                                                const day = (
+                                                                    <td key={"td_" + cell} className="calendar-cell">
+                                                                        <div className="day">
+                                                                            <button
+                                                                                type="button"
+                                                                                className={classNames([
+                                                                                    "day-btn", // prettier
+                                                                                    isSelected &&
```
alexander-rassanov: As a suggestion I would move out this conditions to somewhere to make cleaner code and more maintainable

src/forms/datepicker/DatePicker.tsx
```diff
@@ -0,0 +1,423 @@
...
+                                                                                        !isToday &&
+                                                                                        "is-selected",
+                                                                                    isToday && "is-today",
+                                                                                    isDisabled && "is-disabled",
+                                                                                ])}
+                                                                                tabIndex={
+                                                                                    selectedSameMonth
+                                                                                        ? isSelected
+                                                                                            ? 0
```
alexander-rassanov: let's move out from here this condition

src/forms/datepicker/DatePicker.tsx
```diff
@@ -0,0 +1,423 @@
...
+                                                                return day;
+                                                            })}
+                                                        </tr>
+                                                    );
+                                                })}
+                                            </tbody>
+                                        </table>
+                                    </div>
+                                )}
+                                {viewMode === ViewMode.Month && (
```
alexander-rassanov: You need to render different views for different viewMode let's split them on components/render functions

src/forms/datepicker/DatePicker.tsx
```diff
@@ -0,0 +1,423 @@
...
+                                        {[0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11].map((month: number) => {
+                                            return (
+                                                <button
+                                                    type="button" //prettier
+                                                    className={classNames([
+                                                        "calendar-btn", //prettier
+                                                        "month",
+                                                        month === 0 && "is-selected",
+                                                    ])}
+                                                    autoFocus={month === 0 ? true : false}
```
alexander-rassanov: can be replaced on !month. But it doesn't work if you need number only

src/forms/datepicker/DatePicker.tsx
```diff
@@ -0,0 +1,423 @@
...
+                            type="button"
+                            title="Open"
+                            onClick={this.handleToggle.bind(this)}
+                        >
+                            <Icon shape="calendar" />
+                        </button>
+                        {isOpen && (
+                            <div
+                                className="datepicker"
+                                style={{
```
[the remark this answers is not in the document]
ben-schumacher: @alexander-rassanov I'm not familiar with how to supply inline className value instead of style tag.  This is corner case to get the div working correctly.  Probably not a huge deal.  What do you think?

src/forms/datepicker/DatePicker.tsx
ben-schumacher: Yeah, we need to keep as number.

src/forms/datepicker/DatePicker.tsx
ben-schumacher: Can you explain in more detail please?

src/forms/datepicker/DatePicker.tsx
ben-schumacher: I did some refactoring here.  What do you think now?

src/forms/datepicker/DatePicker.tsx
alexander-rassanov: We will keep components that should be rendered for specific viewMode separately. For example: we move whole <div className="monthpicker"> to another component? As a result we will render the component.
What do you think? Do we need keep components smaller?

src/forms/datepicker/DatePicker.tsx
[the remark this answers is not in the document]
alexander-rassanov: probably we can omit for right now and will discuss later how we will deal with cases like this

src/forms/datepicker/DatePicker.tsx
ben-schumacher: @alexander-rassanov I think for now it would be okay to keep the code together in a single component.  Maybe when time allows, we can do some refactoring.  Because `clarity-react` is a toolkit to be consumed by other applications, the Datepicker component will still be easily consumed/used without this change.  For now, lets keep the same if you're okay with that.

PR #13, opened 2019-06-04
Fix vertical nav icons, removed unneeded clarity package and removed …

ben-schumacher:
…extra index file.


ben-schumacher:
@akate1 That's weird.  I think I deleted and recreated a file with uppercase (I) for Index.ts and modified it to be ./Badges because it was originally ./badges (which was wrong).  I then renamed the file with lowercase (i) index.ts.  If you look at the branch, the file is in fact there: https://github.com/EMCECS/clarity-react/tree/feature-FLEX-513-cleanup-effort/src/emphasis/badges

PR #14, opened 2019-06-04
Enhancing lable component to allow custom styling.

ben-schumacher:


ben-schumacher:
@akate1 Good idea, I added className property.

PR #16, opened 2019-06-12
Added new Progress Bar component

akate1:
FLEX-358 : Port Progress Bars component to React

**Test Details:**
Some screen shots from storybook deployed locally in branch feature-FLEX-358-ProgressBars

**Progress Bar with different colors.**
image

**Labled Progress Bar**
image


src/progress_bars/ProgressBars.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+export class ProgressBar extends React.PureComponent<ProgressBarProps> {
+
+    // By default Progress Bar will be normal progress bar not static
+    static defaultProps = {
+        type:ProgressBarType.NORMAL
+    }
+
+    render() {
+        const {value, max, status, type, labeled, style, className, animation} = this.props;
```
kulakshay: is this line crossing 80 chars?
if so please split each prop on new line

src/progress_bars/ProgressBars.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+        // Label style for labled progress bar
+        const labelStyle = { display: "block"}
+
+        let labeledClassName;
+        if (labeled){
```
kulakshay: Add extra blank line after all variables are declared.

src/progress_bars/ProgressBars.tsx
```diff
@@ -0,0 +1,104 @@
...
+            labeledClassName = 'labeled'
+        }
+
+        let animationClasses;
```
kulakshay: All variables should be declared at the beginning.

akate1: done

src/progress_bars/ProgressBars.tsx
akate1: done

src/progress_bars/ProgressBars.tsx
akate1: done

src/progress_bars/ProgressBars.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+    render() {
+        const {value,
+               max,
+               status,
+               type,
+               labeled,
+               style,
+               className,
+               animation} = this.props;
```
kulakshay: Ideally, it should be formatted like,
```
const {
    propA,
    propB,
    .
    .
} = this.props;
```

src/progress_bars/ProgressBars.tsx
```diff
@@ -0,0 +1,112 @@
...
+                    animationClasses,
+                    className, // prettier
+
+                ])}
+
+                style={style}
+            >
+
+
+                {/* Render normal progress bar if type is normal */}
```
kulakshay: Why comments are in a block like `{}`?

kulakshay:
This is for everyone, 
Since we are doing CRs in github, please do not squash your commits until a complete CR is done.
This helps reviewers to understand the diff between last commit and second last commit.

src/progress_bars/ProgressBars.tsx
akate1: This is how we provide comments in JSX code.

src/progress_bars/ProgressBars.tsx
```diff
@@ -0,0 +1,114 @@
...
+               labeled,
+               style,
+               className,
+               animation
+            } = this.props;
+
+        // Label style for labled progress bar
+        const labelStyle = { display: "block"}
+
+        let Tag: any;
```
ben-schumacher: I think since Tag is *always* equal to "div", we don't need this variable and can explicitly set the JSX element as <div>.

src/progress_bars/ProgressBars.tsx
```diff
@@ -0,0 +1,114 @@
...
+        }
+
+        Tag="div"
+
+        return (
+            <Tag
+                className={classNames([
+                    type,
+                    status,
+                    labeledClassName,
```
ben-schumacher: We can drop this variable and just do: `labeled && "labeled"`

src/progress_bars/ProgressBars.tsx
```diff
@@ -0,0 +1,114 @@
...
+                    animationClasses,
```
ben-schumacher: It might be cleaner to have animation class names come in as a string (comma separated) instead of string[].  That way, you can just drop the property value here and skip the animation.join(' ') above.  That's the nice thing about using the classnames([1, 2, 3]) util function as it will take all comma separated values and format them for you.

ben-schumacher:
> This is for everyone,
> Since we are doing CRs in github, please do not squash your commits until a complete CR is done.
> This helps reviewers to understand the diff between last commit and second last commit.
> 
> Also, before committing the code we should run the eslint, if not already doing so.

I agree, @akate1 can you configure the same way we do this in ecs-flex-ui?  We'll need to add dev dependency lint-staged.

`yarn add lint-staged`

Then, modify packages.json to have a precommit under scripts and then `"lint-staged": {
        "src/**/*.{js,jsx,json,css,ts,tsx}": [
            ".\\node_modules\\.bin\\prettier --write",
            "git add"
        ]
    }`

ben-schumacher:
@akate1,  can you please add another example in storybook that shows using a progress in a card component?

src/progress_bars/ProgressBars.tsx
akate1: removed Tag varriable.

akate1: Removed Tag variable.

src/progress_bars/ProgressBars.tsx
akate1: Removed animation prop and used className prop for same purpose.

akate1:
> @akate1, can you please add another example in storybook that shows using a progress in a card component?
Done.

akate1:
> @akate1, can you please add another example in storybook that shows using a progress in a card component?
Added stories for Progress bar in card component.

ben-schumacher:
@akate1, looks good.  I added one more package (husky) that hooks into git for running prettier.  Seems to work good now.  I also merged latest changes from master into your branch.

akate1:
@ben-schumacher  Thanks for review.

PR #17, opened 2019-06-17
Changes for outlined input.

psivakumaran:
Added style in the props for input box.

### DIFF
--- a/src/forms/input/index.tsx
+++ b/src/forms/input/index.tsx
@@ -14,7 +14,6 @@ import {ReactNode} from "react";
 
 type InputProps = {
     className?: string
-    type?: string
     disabled?: boolean
     helperText?: ReactNode
     label?: string
--- a/src/forms/input/index.tsx
+++ b/src/forms/input/index.tsx
@@ -18,10 +18,12 @@ type InputProps = {
     disabled?: boolean
     helperText?: ReactNode
     label?: string
+    type?: string
     onChange?: (evt: React.ChangeEvent<HTMLInputElement>) => void
     placeholder?: string
     name: string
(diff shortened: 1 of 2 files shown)


psivakumaran:
input_outlined

psivakumaran:
input_outlined

src/forms/input/index.tsx
```diff
@@ -18,10 +18,12 @@ type InputProps = {
     disabled?: boolean
     helperText?: ReactNode
     label?: string
+    type?: string
```
akate1: I think 'type' prop is already present at line 17 in same file.

src/forms/input/index.tsx
```diff
@@ -55,11 +57,16 @@ export class Input extends React.PureComponent<InputProps> {
...
-            type
+            type,
+            style
         } = this.props;
         let classNames = ["clr-control-container", className];
+        let styled: any;
         if (disabled)
             classNames.push("clr-form-control-disabled");
+        styled = {
+             ...style
```
akate1: I guess we can directly use 'style' prop as style = {style}.

src/forms/input/index.tsx
psivakumaran: removed the duplicate

src/forms/input/index.tsx
psivakumaran: Yeah I thought of doing the same way, but I kept in this format in case we need to add other options to styling as we have done for label.

ben-schumacher:
@psivakumaran, going to close this PR as the issue was resolved in another PR.  Thanks!

PR #20, opened 2019-06-25
Adding support for input icons with additional properties and render …

ben-schumacher:
…logic.


src/forms/input/index.tsx
```diff
@@ -47,7 +47,17 @@ export class Input extends React.PureComponent<InputProps> {
...
+        const {
+            className, //prettier
+            disabled,
+            helperText,
+            label,
+            isBoxed,
+            placeholder,
+            size,
+            type,
+            children,
```
akate1: I think we also need to add  prop named as 'children' in InputProps.
For example ,children?: ReactNode | ReactNode[];

ben-schumacher: @akate1, this is not required as `children` is a property always available in React components.

akate1: ok got it.

girishdudhe:
Why the name of file is index.tsx? It should be Input.tsx
Is there any specific reason?

ben-schumacher:
> Why the name of file is index.tsx? It should be Input.tsx
> Is there any specific reason?

I agree, renamed.  Thanks @girishdudhe !

PR #24, opened 2019-07-05
[FLEX-533] changes for adding classname to CardBlock

psivakumaran:


src/cards/Card.tsx
```diff
@@ -16,8 +16,11 @@ export const CardImage: React.FunctionComponent = ({children}) => {
     return <div className={ClassNames.CARD_IMAGE}>{children}</div>;
 };
 
-export const CardBlock: React.FunctionComponent = ({children}) => {
-    return <div className={ClassNames.CARD_BLOCK}>{children}</div>;
+export type CardBlockProps = {
```
akate1: You don't need to export CardBlockProps. As currently you are using them in same file only.

ben-schumacher: I agree here.

psivakumaran: Done.

psivakumaran:
> @psivakumaran I think the changes look fine, but just wondering why you need to set custom className on CardBlock? Can you provide some context please?

I have added these to give a grey background to the Objectstore name and menu section.
When I directly use div/span, there is a white border around the div which is part of the card block

PR #27, opened 2019-07-10
[FELX-659]: Adding changes for login page's component

girishdudhe:
1. Adding className prop for Alert
2. Adding value prop for Input

### DIFF
--- a/src/forms/input/Input.tsx
+++ b/src/forms/input/Input.tsx
@@ -22,6 +22,7 @@ type InputProps = {
     onChange?: (evt: React.ChangeEvent<HTMLInputElement>) => void;
     placeholder?: string;
     name: string;
+    value?: string;
     size?: number;
 };
 
@@ -52,6 +53,7 @@ export class Input extends React.PureComponent<InputProps> {
             disabled,
             helperText,
             label,
+            value,
             isBoxed,
             placeholder,
             size,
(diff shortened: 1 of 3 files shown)


src/forms/input/Input.tsx
```diff
@@ -81,6 +84,7 @@ export class Input extends React.PureComponent<InputProps> {
                         <input
                             type={type || "text"}
                             id={name}
+                            value={value}
```
akate1: Can we add story with new 'value' prop in storybook?

girishdudhe: Addressed...

ben-schumacher:
@akate1 Can you please review one more time and approve if everything looks good?  Thank you!

src/forms/input/Input.tsx
```diff
@@ -66,6 +70,8 @@ export class Input extends React.PureComponent<InputProps> {
                 <input
                     type={type || "text"}
                     id={name}
+                    value={value}
+                    defaultValue={defaultValue}
```
akate1: As per my knowledge you can not set 'defaultValue' directly to input HTML tag. 

You can set 'defaultValue' programmatically once, and then let user do whatever.With uncontrolled values, the value attribute (value={value here}) inside the input element will override any values that are typed in so in order to work around that, React/JavaScript provides another attribute called defaultValue (defaultValue={value here}) that will pre-populate the input field with the defaultValue without overriding any value input by the user.

kulakshay: `defaultValue` prop is supported by React, you can read more about it https://reactjs.org/docs/uncontrolled-components.html.
In this case, it might not be required though.

akate1: @kulakshay  Thanks for clarifying this.

PR #30, opened 2019-07-17
[FLEX 522] add numeric +/- button in input

girishdudhe:
1. Adding the number +/- in input
2. Adding onClick event to Icon

I added the Error Message in label that is not mentioned in JIRA ticket. As this message is directly on UI so it should be i18N format so I have provided errMsg prop to InputCounter. 
Storybook for InputCounter:
http://10.86.78.152:6006/?path=/story/inputcounter--a-simple-inputcounter-box

### DIFF
--- a/src/forms/inputcounter/InputCounter.tsx
+++ b/src/forms/inputcounter/InputCounter.tsx
@@ -11,13 +11,13 @@ import * as React from "react";
 import * as utils from "../../utils";
 import {Icon} from "../../icon";
 import {Input} from "../input/Input";
+import "./InputCounter.css";
 
 type InputCounterProps = {
     className?: string;
     maxValue?: number;
     minValue?: number;
     defaultValue?: number;
-    name?: string;
     errMsg?: string;
     width?: string;
     onChange?: (evt: React.ChangeEvent<HTMLInputElement>) => void;
@@ -32,36 +32,33 @@ export class InputCounter extends React.PureComponent<InputCounterProps, InputCo
     public static defaultProps: Partial<InputCounterProps> = {
         minValue: Number.MIN_SAFE_INTEGER,
         maxValue: Number.MAX_SAFE_INTEGER,
+        errMsg: "Value should be",
+        width: "6rem",
     };
     constructor(props: any) {
         super(props);
 
+        //Default Value should be >= to minValue
         let defaultValue = isNaN(props.defaultValue) ? 0 : Math.max(props.defaultValue, props.minValue);
+        //Default Value should be <= to minValue
+        defaultValue = Math.min(defaultValue, props.maxValue);
         this.state = {
             value: defaultValue,
             errorMsg: "",
         };
     }
 
-    private getErrorMessage() {
-        if (this.props.errMsg === undefined) {
-            return "Value should be ";
-        } else {
-            return this.props.errMsg;
-        }
-    }
-
     private handleChange = (evt: React.ChangeEvent<HTMLInputElement>) => {
         if (isNaN(parseInt(evt.target.value))) {
             this.setState({
-                value: isNaN(this.props.minValue!) ? 0 : this.props.minValue!,
+                value: isNaN(this.props.minValue!) ? 0 : Math.max(this.props.defaultValue, this.props.minValue!),
             });
         } else if (parseInt(evt.target.value) > this.props.maxValue!) {
-            let maxValueError = this.getErrorMessage() + " <= " + this.props.maxValue;
+            let maxValueError = this.props.errMsg + " <= " + this.props.maxValue;
             this.setState({errorMsg: maxValueError});
             this.setState({value: parseInt(evt.target.value)});
         } else if (parseInt(evt.target.value) < this.props.minValue!) {
-            let minValueError = this.getErrorMessage() + " >= " + this.props.minValue;
+            let minValueError = this.props.errMsg + " >= " + this.props.minValue;
             this.setState({errorMsg: minValueError});
             this.setState({value: parseInt(evt.target.value)});
         } else {
@@ -73,56 +70,40 @@ export class InputCounter extends React.PureComponent<InputCounterProps, InputCo
     };
 
     private handleIncrement() {
-        if (this.props.maxValue === undefined) {
-            this.setState({value: this.state.value + 1});
-        } else if (this.state.value < this.props.maxValue!) {
+        if (this.state.value < this.props.maxValue!) {
             this.setState({value: this.state.value + 1});
         } else {
-            let maxValueError = this.getErrorMessage() + " <= " + this.props.maxValue;
+            let maxValueError = this.props.errMsg + " <= " + this.props.maxValue;
             this.setState({errorMsg: maxValueError});
         }
     }
 
     private handleDecrement() {
-        if (this.props.minValue === undefined) {
-            this.setState({value: this.state.value - 1});
-        } else if (this.state.value > this.props.minValue!) {
+        if (this.state.value > this.props.minValue!) {
             this.setState({value: this.state.value - 1});
         } else {
-            let minValueError = this.getErrorMessage() + " >= " + this.props.minValue;
+            let minValueError = this.props.errMsg + " > " + this.props.minValue;
             this.setState({errorMsg: minValueError});
         }
     }
 
     render() {
-        const {maxValue, minValue, defaultValue, onChange, name, errMsg} = this.props;
-        let {width} = this.props;
-        if (width === undefined) {
-            width = "6rem"; //This is defaultWidth
-        }
+        const {width, maxValue, minValue, defaultValue, onChange, errMsg} = this.props;
         return (
             <div className="clr-form" style={{width: width}}>
-                <div style={{display: "flex"}}>
-                    <div style={{display: "inline-block", paddingRight: "0%", paddingLeft: "0%"}}>
+                <div className="ic-flex">
+                    <div className="ic-div">
                         <input
+                            className="ic-input"
                             name="counter"
                             value={this.state.value}
                             onChange={this.handleChange.bind(this)}
-                            style={{height: "52px", width: "100%", border: "1px solid #d3d3d3"}}
                         />
                     </div>
-                    <div style={{display: "inline-block"}}>
-                        <Icon
-                            shape="plus"
-                            style={{height: "26px", border: "1px solid #d3d3d3", paddingRight: "0%", paddingLeft: "0%"}}
-                            onClick={this.handleIncrement.bind(this)}
-                        />
+                    <div className="ic-div">
+                        <Icon className="ic-icon" shape="plus" onClick={this.handleIncrement.bind(this)} />
                         <br />
-                        <Icon
-                            shape="minus"
-                            style={{height: "26px", border: "1px solid #d3d3d3", paddingRight: "0%", paddingLeft: "0%"}}
-                            onClick={this.handleDecrement.bind(this)}
-                        />
+                        <Icon className="ic-icon" shape="minus" onClick={this.handleDecrement.bind(this)} />
                     </div>
                 </div>
                 <div>
--- a/src/forms/inputcounter/InputCounter.tsx
+++ b/src/forms/inputcounter/InputCounter.tsx
@@ -29,7 +29,7 @@ type InputCounterState = {
 };
 
 export class InputCounter extends React.PureComponent<InputCounterProps, InputCounterState> {
-    public static defaultProps: Partial<InputCounterProps> = {
+    static defaultProps = {
         minValue: Number.MIN_SAFE_INTEGER,
         maxValue: Number.MAX_SAFE_INTEGER,
         errMsg: "Value should be",
@@ -51,7 +51,7 @@ export class InputCounter extends React.PureComponent<InputCounterProps, InputCo
     private handleChange = (evt: React.ChangeEvent<HTMLInputElement>) => {
         if (isNaN(parseInt(evt.target.value))) {
             this.setState({
-                value: isNaN(this.props.minValue!) ? 0 : Math.max(this.props.defaultValue, this.props.minValue!),
+                value: isNaN(this.props.minValue!) ? 0 : Math.max(this.props.defaultValue!, this.props.minValue!),
             });
         } else if (parseInt(evt.target.value) > this.props.maxValue!) {
             let maxValueError = this.props.errMsg + " <= " + this.props.maxValue;
--- a/src/forms/inputcounter/InputCounter.tsx
+++ b/src/forms/inputcounter/InputCounter.tsx
@@ -1,0 +1,131 @@
+/**
+ * Copyright (c) 2018 Dell Inc., or its subsidiaries. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ */
+import * as React from "react";
+import * as utils from "../../utils";
+import {ReactNode} from "react";
+import {Button} from "../button";
+import {Input} from "../input/Input";
+
+type InputCounterProps = {
+    className?: string;
+    maxValue: string;
+    minValue: string;
+    defaultValue?: string | "0";
+    onChange?: (evt: React.ChangeEvent<HTMLInputElement>) => void;
+    name?: string;
+    style?: any;
+};
+
+type InputCounterState = {
+    value: number;
+    errorMsg: string;
+};
+
+export class InputCounter extends React.PureComponent<InputCounterProps, InputCounterState> {
+    constructor(props: any) {
+        super(props);
+        this.state = {
+            value: parseInt(props.defaultValue),
+            errorMsg: "",
+        };
+    }
+
+    private handleChange = (evt: React.ChangeEvent<HTMLInputElement>) => {
+        if (isNaN(parseInt(evt.target.value))) {
+            this.setState({value: parseInt(this.props.minValue)});
+        } else if (parseInt(evt.target.value) > parseInt(this.props.maxValue)) {
+            let maxValueError = "Value should be < " + this.props.maxValue;
+            this.setState({errorMsg: maxValueError});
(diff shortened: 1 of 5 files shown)


src/forms/inputcounter/InputCounter.tsx
```diff
@@ -0,0 +1,134 @@
...
+        let defaultValue = isNaN(props.defaultValue) ? 0 : Math.max(props.defaultValue, props.minValue);
+        this.state = {
+            value: defaultValue,
+            errorMsg: "",
+        };
+    }
+
+    private getErrorMessage() {
+        if (this.props.errMsg === undefined) {
+            return "Value should be ";
```
akate1: You can add this error message  in default props.

src/forms/inputcounter/InputCounter.tsx
```diff
@@ -0,0 +1,134 @@
...
+        }
+        return (
+            <div className="clr-form" style={{width: width}}>
+                <div style={{display: "flex"}}>
+                    <div style={{display: "inline-block", paddingRight: "0%", paddingLeft: "0%"}}>
+                        <input
+                            name="counter"
+                            value={this.state.value}
+                            onChange={this.handleChange.bind(this)}
+                            style={{height: "52px", width: "100%", border: "1px solid #d3d3d3"}}
```
akate1: Try specify style in uniform unit all should be in form of % or in term of 'px'

src/forms/inputcounter/InputCounter.tsx
```diff
@@ -0,0 +1,134 @@
...
+        super(props);
+
+        let defaultValue = isNaN(props.defaultValue) ? 0 : Math.max(props.defaultValue, props.minValue);
+        this.state = {
+            value: defaultValue,
+            errorMsg: "",
+        };
+    }
+
+    private getErrorMessage() {
```
akate1: You can remove this function by providing default value to 'errMsg' prop

src/forms/inputcounter/InputCounter.tsx
```diff
@@ -0,0 +1,134 @@
...
+    render() {
+        const {maxValue, minValue, defaultValue, onChange, name, errMsg} = this.props;
+        let {width} = this.props;
+        if (width === undefined) {
+            width = "6rem"; //This is defaultWidth
+        }
+        return (
+            <div className="clr-form" style={{width: width}}>
+                <div style={{display: "flex"}}>
+                    <div style={{display: "inline-block", paddingRight: "0%", paddingLeft: "0%"}}>
```
psivakumaran: The styles can be added to a css file.

src/forms/inputcounter/InputCounter.tsx
psivakumaran: Same here, the styles can be added to a css file.

src/forms/inputcounter/InputCounter.tsx
```diff
@@ -0,0 +1,134 @@
...
+                            name="counter"
+                            value={this.state.value}
+                            onChange={this.handleChange.bind(this)}
+                            style={{height: "52px", width: "100%", border: "1px solid #d3d3d3"}}
+                        />
+                    </div>
+                    <div style={{display: "inline-block"}}>
+                        <Icon
+                            shape="plus"
+                            style={{height: "26px", border: "1px solid #d3d3d3", paddingRight: "0%", paddingLeft: "0%"}}
```
psivakumaran: Same, the styles can be added to a css file.

src/forms/inputcounter/InputCounter.tsx
```diff
@@ -0,0 +1,134 @@
...
+                            onClick={this.handleIncrement.bind(this)}
+                        />
+                        <br />
+                        <Icon
+                            shape="minus"
+                            style={{height: "26px", border: "1px solid #d3d3d3", paddingRight: "0%", paddingLeft: "0%"}}
```
psivakumaran: Same, the styles can be added to a css file.

src/forms/inputcounter/InputCounter.tsx
```diff
@@ -0,0 +1,134 @@
...
+    defaultValue?: number;
+    name?: string;
+    errMsg?: string;
+    width?: string;
+    onChange?: (evt: React.ChangeEvent<HTMLInputElement>) => void;
+};
+
+type InputCounterState = {
+    value: number;
+    errorMsg: string;
```
akate1: Instate of keeping error message in state can we derive it at runtime ?

src/forms/inputcounter/InputCounter.tsx
```diff
@@ -0,0 +1,134 @@
...
+import * as utils from "../../utils";
+import {Icon} from "../../icon";
+import {Input} from "../input/Input";
+
+type InputCounterProps = {
+    className?: string;
+    maxValue?: number;
+    minValue?: number;
+    defaultValue?: number;
+    name?: string;
```
akate1: Not getting use of 'name' prop here.

src/forms/inputcounter/InputCounter.tsx
```diff
@@ -0,0 +1,134 @@
...
+    errMsg?: string;
+    width?: string;
```
akate1: instate of only width it is better if we give control of whole style by using 'style' prop.

src/forms/inputcounter/InputCounter.tsx
```diff
@@ -0,0 +1,134 @@
...
+    public static defaultProps: Partial<InputCounterProps> = {
+        minValue: Number.MIN_SAFE_INTEGER,
+        maxValue: Number.MAX_SAFE_INTEGER,
+    };
+    constructor(props: any) {
+        super(props);
+
+        let defaultValue = isNaN(props.defaultValue) ? 0 : Math.max(props.defaultValue, props.minValue);
+        this.state = {
+            value: defaultValue,
```
akate1: No need of defaultValue variable you can directly replace it with 'isNaN(props.defaultValue) ? 0 : Math.max(props.defaultValue, props.minValue);' here

src/forms/inputcounter/InputCounter.tsx
```diff
@@ -0,0 +1,134 @@
...
+            return "Value should be ";
+        } else {
+            return this.props.errMsg;
+        }
+    }
+
+    private handleChange = (evt: React.ChangeEvent<HTMLInputElement>) => {
+        if (isNaN(parseInt(evt.target.value))) {
+            this.setState({
+                value: isNaN(this.props.minValue!) ? 0 : this.props.minValue!,
```
psivakumaran: Should this be set to defaultValue after comparison with minVal and maxVal? Like in constructor.

src/forms/inputcounter/InputCounter.tsx
payald: The text can be added as a constant

src/forms/inputcounter/InputCounter.tsx
```diff
@@ -0,0 +1,134 @@
...
+        } else {
+            this.setState({value: parseInt(evt.target.value)});
+            this.setState({errorMsg: ""});
+        }
+        const {onChange} = this.props;
+        if (onChange) onChange(evt);
+    };
+
+    private handleIncrement() {
+        if (this.props.maxValue === undefined) {
```
akate1: You have already define default value for 'maxValue' then it can not be undefined here.

src/forms/inputcounter/InputCounter.tsx
```diff
@@ -0,0 +1,134 @@
...
+            this.setState({value: this.state.value + 1});
+        } else if (this.state.value < this.props.maxValue!) {
```
akate1: What if 'value' is equal to 'maxValue' ?

src/forms/inputcounter/InputCounter.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+export class InputCounter extends React.PureComponent<InputCounterProps, InputCounterState> {
+    public static defaultProps: Partial<InputCounterProps> = {
+        minValue: Number.MIN_SAFE_INTEGER,
+        maxValue: Number.MAX_SAFE_INTEGER,
+    };
+    constructor(props: any) {
+        super(props);
+
+        let defaultValue = isNaN(props.defaultValue) ? 0 : Math.max(props.defaultValue, props.minValue);
```
psivakumaran: Should we also compare with maxValue? What happens if the defaultValue is greater that maxVal?

src/forms/inputcounter/InputCounter.tsx
```diff
@@ -0,0 +1,134 @@
...
+        } else if (this.state.value < this.props.maxValue!) {
+            this.setState({value: this.state.value + 1});
+        } else {
+            let maxValueError = this.getErrorMessage() + " <= " + this.props.maxValue;
+            this.setState({errorMsg: maxValueError});
+        }
+    }
+
+    private handleDecrement() {
+        if (this.props.minValue === undefined) {
```
akate1: You have already define default value for 'minValue' then it can not be undefined here.

src/forms/inputcounter/InputCounter.tsx
```diff
@@ -0,0 +1,134 @@
...
+            this.setState({value: this.state.value - 1});
+        } else if (this.state.value > this.props.minValue!) {
```
akate1: What if 'value' is equal to 'minValue' ? We need to handle this case as well

src/forms/inputcounter/InputCounter.tsx
girishdudhe: I just updated the message in commit message for this propr errMsg. If errMsg prop is approved by everyone then I will modify it as constant.

src/forms/inputcounter/InputCounter.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+    private getErrorMessage() {
+        if (this.props.errMsg === undefined) {
+            return "Value should be ";
+        } else {
+            return this.props.errMsg;
+        }
+    }
+
+    private handleChange = (evt: React.ChangeEvent<HTMLInputElement>) => {
```
akate1: Are we allowing user to directly edit value in textbox ?

src/forms/inputcounter/InputCounter.tsx
```diff
@@ -0,0 +1,134 @@
...
+                            value={this.state.value}
+                            onChange={this.handleChange.bind(this)}
+                            style={{height: "52px", width: "100%", border: "1px solid #d3d3d3"}}
+                        />
+                    </div>
+                    <div style={{display: "inline-block"}}>
+                        <Icon
+                            shape="plus"
+                            style={{height: "26px", border: "1px solid #d3d3d3", paddingRight: "0%", paddingLeft: "0%"}}
+                            onClick={this.handleIncrement.bind(this)}
```
akate1: Since this component do not have specific CSS classes like Toast can we write CSS classes for this rather than inline styles ?

src/forms/inputcounter/InputCounter.tsx
girishdudhe: As we have given only width as actual prop that is going to change. So added % only for width. All other are added as px as those are not going to set from outside.

src/forms/inputcounter/InputCounter.tsx
girishdudhe: yes

src/forms/inputcounter/InputCounter.tsx
girishdudhe: Actually is there any example of css class in clarity-react repo....I have not seen .css file so added all as inline.

src/forms/inputcounter/InputCounter.tsx
girishdudhe: Addressed

src/forms/inputcounter/InputCounter.tsx
girishdudhe: ok got it....Addressed also removed getErrorMessage as it is not required now.

src/forms/inputcounter/InputCounter.tsx
girishdudhe: done

src/forms/inputcounter/InputCounter.tsx
girishdudhe: what is the problem in state? If we remove it from state, then we need to use ref that add complexity in code.

src/forms/inputcounter/InputCounter.tsx
girishdudhe: ok. Addressed.

src/forms/inputcounter/InputCounter.tsx
girishdudhe: We should not allow to increment value and show the error message.

src/forms/inputcounter/InputCounter.tsx
girishdudhe: Addressed this in both the handler.

src/forms/inputcounter/InputCounter.tsx
girishdudhe: It is handled here. If it is equal, then don't allow to decrements the value and show the error message.

src/forms/inputcounter/InputCounter.tsx
girishdudhe: Added the width also in default Props.

girishdudhe: Removed this function....

src/forms/inputcounter/InputCounter.tsx
girishdudhe: Addressed.

src/forms/inputcounter/InputCounter.tsx
girishdudhe: Addressed.

src/forms/inputcounter/InputCounter.tsx
girishdudhe: Addressed.

src/forms/inputcounter/InputCounter.tsx
girishdudhe: Addressed.

src/forms/inputcounter/InputCounter.tsx
girishdudhe: Addressed.

src/forms/inputcounter/InputCounter.tsx
girishdudhe: As I need to allow only width to change, I have not give style as prop.

src/forms/inputcounter/InputCounter.tsx
```diff
@@ -0,0 +1,115 @@
...
+    onChange?: (evt: React.ChangeEvent<HTMLInputElement>) => void;
+};
+
+type InputCounterState = {
+    value: number;
+    errorMsg: string;
+};
+
+export class InputCounter extends React.PureComponent<InputCounterProps, InputCounterState> {
+    public static defaultProps: Partial<InputCounterProps> = {
```
akate1: you can remove public keyword here. Also not getting what is use of 'Partial' here ?

src/forms/inputcounter/InputCounter.tsx
```diff
@@ -0,0 +1,115 @@
...
+        errMsg: "Value should be",
+        width: "6rem",
+    };
+    constructor(props: any) {
+        super(props);
+
+        //Default Value should be >= to minValue
+        let defaultValue = isNaN(props.defaultValue) ? 0 : Math.max(props.defaultValue, props.minValue);
+        //Default Value should be <= to minValue
+        defaultValue = Math.min(defaultValue, props.maxValue);
```
akate1: What is purpose of changing defaultValue here again ?  I guess defaultValue is given by user and you just need to check if it is between Min and Max value. Can't we do this in one check? And what if user provide wrong default value ?

src/forms/inputcounter/InputCounter.tsx
```diff
@@ -0,0 +1,115 @@
...
+        } else {
+            this.setState({value: parseInt(evt.target.value)});
+            this.setState({errorMsg: ""});
+        }
+        const {onChange} = this.props;
+        if (onChange) onChange(evt);
+    };
+
+    private handleIncrement() {
+        if (this.state.value < this.props.maxValue!) {
```
akate1: Same logic is repeated in handleChange() to check validity of value before increment/decrements. Can we make common function of it and reuse at both places ?

src/forms/inputcounter/InputCounter.tsx
girishdudhe: Addressed.

src/forms/inputcounter/InputCounter.tsx
akate1: There is no problem here. But  just a thought how does error message will affect state of this component? .Does this is something that component want to remember throughout state change and  does next state depends on error message?

src/forms/inputcounter/InputCounter.tsx
girishdudhe: Login is different in both the places. In handleIncrement(), it is checking against maxValue and in handleDecrement(). it is checking against minValue.

akate1: Yes I know.  I just combined 2 comments here. We can create 2 common methods and  reuse them.

src/forms/inputcounter/InputCounter.tsx
girishdudhe: DefaultValue should be >= MinValue and  <= MaxValue. So even if user gives the defaultValue < MinValue or defaultValue>maxValue then we need to adjust the defaultValue as per minValue and maxValue.
So to address the comment given by priya: "What happens if the defaultValue is greater that maxVal?" , I have added the code to select the min value between defaultValue and maxValue.

akate1: @girishdudhe  Its not wise to change value without knowledge of user. We should give some error. I am not sure but please check if we  can add such conditions while declaration of props itself

src/forms/inputcounter/InputCounter.tsx
girishdudhe: This "Partial<Props>" is used to assign value to partial props as default value in typescript version < 3. As we are using typescript version > 3, we can remove that.

girishdudhe:
We are going to use the solution suggested by Ben.

PR #31, opened 2019-07-17
FLEX-638 Port Wizard component to React

akate1:
**Description:**

This is PR for porting wizard component to React. 
Link: https://clarity.design/documentation/wizards#examples

This PR has covered following tasks.
- Added className prop in NAV, NAVLINK and VerticalNav component
- Render basic wizard
- Added story for wizard with different sizes
- Added story for opening wizard at predefined step
- Add story for overriding wizard buttons
- Add story for overriding wizard without navigation links
- Solved CSS issue on hover of wizard nav link
- Add logic to reset wizard and story for same
- Add logic for asynchronous and synchronous validation
- Add story for synchronous and asynchronous validation with wizard

image

### DIFF
--- a/src/layout/nav/Nav.tsx
+++ b/src/layout/nav/Nav.tsx
@@ -19,7 +19,6 @@ export type NavProps = {
 
     // The navigation type (header, sub side).
     navType: NavType;
-    className?: string;
 };
 
 export enum NavLevel {
@@ -64,8 +63,8 @@ export class Nav extends React.PureComponent<NavProps> {
     }
 
     render() {
-        const {children, navLevel, navType, className} = this.props;
-        const classList: string[] = [Nav.classForNavType(navType), Nav.classForNavLevel(navLevel), className as any];
+        const {children, navLevel, navType} = this.props;
+        const classList: string[] = [Nav.classForNavType(navType), Nav.classForNavLevel(navLevel)];
         return <nav className={utils.classNames(classList)}>{Nav.maybeWrapInList(navType, children)}</nav>;
     }
 }
--- a/src/layout/nav/Nav.tsx
+++ b/src/layout/nav/Nav.tsx
@@ -19,6 +19,7 @@ export type NavProps = {
 
     // The navigation type (header, sub side).
     navType: NavType;
+    className?: string;
 };
 
 export enum NavLevel {
@@ -63,8 +64,8 @@ export class Nav extends React.PureComponent<NavProps> {
     }
 
     render() {
-        const {children, navLevel, navType} = this.props;
-        const classList: string[] = [Nav.classForNavType(navType), Nav.classForNavLevel(navLevel)];
+        const {children, navLevel, navType, className} = this.props;
+        const classList: string[] = [Nav.classForNavType(navType), Nav.classForNavLevel(navLevel), className as any];
         return <nav className={utils.classNames(classList)}>{Nav.maybeWrapInList(navType, children)}</nav>;
     }
 }
--- a/src/layout/nav/NavLink.tsx
+++ b/src/layout/nav/NavLink.tsx
@@ -27,7 +27,7 @@ export class NavLink extends React.PureComponent<NavLinkProps> {
     render() {
         const {iconShape, className, type, onClick, children} = this.props;
 
-        let cssClass = classNames(["nav-text", type]);
+        let navLinkClassNames = classNames(["nav-text", type]);
         return iconShape ? (
             <a
                 href="javascript:void(0)"
@@ -35,7 +35,7 @@ export class NavLink extends React.PureComponent<NavLinkProps> {
                 onClick={onClick}
             >
                 <Icon shape={iconShape} className="nav-icon" />
-                <span className={cssClass}>{children}</span>
+                <span className={navLinkClassNames}>{children}</span>
             </a>
         ) : (
             <a
@@ -43,7 +43,7 @@ export class NavLink extends React.PureComponent<NavLinkProps> {
                 className={classNames([className, type == NavLinkType.stepNavLink ? "" : "nav-link"])}
                 onClick={onClick}
             >
-                <div className={cssClass}>{children}</div>
+                <div className={navLinkClassNames}>{children}</div>
             </a>
         );
     }
--- a/src/layout/nav/NavLink.tsx
+++ b/src/layout/nav/NavLink.tsx
@@ -15,19 +15,35 @@ import {classNames} from "../../utils";
 export type NavLinkProps = {
     iconShape?: string;
     className?: string;
+    type?: NavLinkType;
+    onClick?: (evt: React.MouseEvent<HTMLElement>) => void;
 };
+
+export enum NavLinkType {
+    stepNavLink = "btn btn-link clr-wizard-stepnav-link",
+}
 
 export class NavLink extends React.PureComponent<NavLinkProps> {
     render() {
-        const {iconShape, className, children} = this.props;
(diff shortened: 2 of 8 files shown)


src/wizards/Wizard.tsx
```diff
@@ -0,0 +1,380 @@
...
+                                        {/*Close modal-body*/}
+                                        {this.buildWizardFooter()}
+                                    </div>{" "}
+                                    {/*Close modal-content*/}
+                                </div>{" "}
+                                {/*Close modal-content-wrapper */}
+                                <div className={ClassNames.MODAL_GHOST_WRAPPER}>
+                                    <div
+                                        _ngcontent-c7=""
+                                        className="modal-ghost modal-ghost-1 ng-trigger ng-trigger-ghostPageOneState"
```
psivakumaran: Is it possible we can move these to the common Classnames file?

src/wizards/Wizard.stories.tsx
```diff
@@ -0,0 +1,133 @@
...
+    {stepName: "page 3", stepId: 2, stepComponent: <p> Page 3</p>, stepCompleted: false},
+];
+
+const stepsLarge = [
+    {stepName: "page 1", stepId: 0, stepComponent: <p> Page 1</p>, stepCompleted: false},
+    {stepName: "page 2", stepId: 1, stepComponent: <p> Page 2</p>, stepCompleted: false},
+    {stepName: "page 3", stepId: 2, stepComponent: <p> Page 3</p>, stepCompleted: false},
+];
+
+const stepsXLarge = [
```
psivakumaran: The variables "stepsLarge", "stepsMedium" and "stepsXLarge" seems to have similar data, can we create a single data structure which can be used for all the stories?

src/wizards/Wizard.tsx
```diff
@@ -0,0 +1,380 @@
...
+    // Close the wizard on finish
+    finishButtonClick() {
+        const {onFinish} = this.props;
+        onFinish && onFinish();
+        this.close();
+    }
+
+    modifyButtonStates(step: WizardStep) {
+        const {steps} = this.props;
+        console.log("in modify button");
```
psivakumaran: Please remove console.log statements.

src/wizards/ClassNames.tsx
```diff
@@ -0,0 +1,39 @@
...
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+export class ClassNames {
+    public static WIZARD: string = "clr-wizard";
+    public static WIZARD_OPEN: string = "ng-tns-c7-2 open";
+    public static WIZARD_MODAL: string = "modal ng-tns-c9-2 ng-star-inserted";
+    public static WIZARD_MODAL_DIALOG: string = "modal-dialog ng-trigger ng-trigger-fadeDown";
```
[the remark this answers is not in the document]
akate1: Yes. We need to use it this way for wizard.

src/wizards/Wizard.stories.tsx
akate1: Though the data is same it is belong to different wizard instance. So need to keep them separate.

src/wizards/Wizard.tsx
```diff
@@ -0,0 +1,380 @@
...
+        showNav: true,
+        defaultStep: 1,
+    };
+
+    // Initial state of wizard
+    state: WizardState = {
+        show: this.props.show !== undefined ? this.props.show : false,
+        currentStep: this.props.defaultStep! - 1 || 0,
+        showFinishButton: false,
+        showPriviousButton: false,
```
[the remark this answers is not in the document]
akate1: Done.Thanks for noticing this

src/wizards/Wizard.tsx
akate1: done

src/layout/nav/NavLink.tsx
```diff
@@ -15,19 +15,35 @@ import {classNames} from "../../utils";
 export type NavLinkProps = {
     iconShape?: string;
     className?: string;
+    type?: NavLinkType;
+    onClick?: (evt: React.MouseEvent<HTMLElement>) => void;
 };
 
+export enum NavLinkType {
+    stepNavLink = "btn btn-link clr-wizard-stepnav-link",
```
ben-schumacher: Do we need to pass this value in from parent component?  Would it be better to just have a property in NavLinkProps?  For example, we could do something like `isWizardStep: boolean`?  Not a blocking comment -just a suggestion.

src/wizards/Wizard.tsx
```diff
@@ -0,0 +1,415 @@
...
+        const {steps, title, showNav} = this.props;
+
+        return (
+            <VerticalNav className={ClassNames.WIZARD_STEPNAV_WRAPPER}>
+                <h3 className={ClassNames.WIZARD_TITLE}> {title} </h3>
+                {showNav && (
+                    <div className={ClassNames.WIZARD_STEPNAV}>
+                        {steps.map((step, key) => {
+                            return (
+                                <NavLink
```
ben-schumacher: When I inspect the html from VMWare Clarity, I can see it's using <button/> instead of <a/>.  It might be more appropriate for us to return something like: `<Button link={true}, icon={step.customStepNav.stepNavIcon} className="clr-wizard-stepnav-link">`  If we do that, we might not need any changes in NavLink.tsx.

src/wizards/Wizard.tsx
```diff
@@ -0,0 +1,415 @@
...
+    stepComponent: React.ReactNode;
+    stepId: number;
+    stepCompleted: boolean;
+    customStepNav?: WizardStepNavDetails;
+    isStepValid?: Function /* This function should return boolen value. And use to determine if step is valid or not */;
+};
+
+export type WizardStepNavDetails = {
+    stepNavIcon?: string;
+    stepNavbadges?: React.ReactNode;
```
ben-schumacher: Maybe name this more general in case we want to use a different component inline.  In case we decided to use something other than badges.

src/layout/nav/Nav.tsx
```diff
@@ -19,6 +19,7 @@ export type NavProps = {
 
     // The navigation type (header, sub side).
     navType: NavType;
+    className?: string;
```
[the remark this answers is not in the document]
akate1: Previously I try to use NAV component instead of VerticalNAV but now I am keeping this for future control.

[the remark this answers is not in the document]
akate1: Previously I try to use NAV component instead of VerticalNAV but now I am keeping this for future control.

src/layout/nav/NavLink.tsx
akate1: @ben-schumacher Nice suggestion. I have change the code accordingly.

src/wizards/Wizard.tsx
akate1: I tried to use our Button component as you suggested but it was causing CSS issues for Wizard Navigation.

src/wizards/Wizard.tsx
akate1: Thanks for noticing that. Changed the variable name

src/wizards/Wizard.tsx
akate1: As discussed changed this code to use 'Button'.

akate1:
> The changes look good to me. Approved.
> Just one comment -
> Do we need to add to index.tsx file to export the modules and required classnames?

We don't want to export classnames outside the Wizard module but I will add index.ts file to keep consistency with other common components.

PR #37, opened 2019-08-14
Changed components to take in icon as IconProps instead of string.

ben-schumacher:


src/wizards/Wizard.tsx
```diff
@@ -387,7 +387,7 @@ export class Wizard extends React.PureComponent<WizardProps> {
                                         onClick={this.navigationClick.bind(this, step.stepId)}
                                         icon={
                                             step.customStepNav !== undefined && step.customStepNav.stepNavIcon
-                                                ? step.customStepNav.stepNavIcon
+                                                ? {shape: step.customStepNav.stepNavIcon}
```
alexander-rassanov: I would avoid nested ?: condition, since it is unreadable. let's create a function for that that will return a result or we can use "if"(above the return statement) and set to a variable that will be use here

PR #39, opened 2019-08-27
New Portlet component for feature-flex-1000

vinodwani18:


src/portlet/Portlet.tsx
```diff
@@ -0,0 +1,94 @@
...
+    state: PortletState = {
+        toggleContent: true,
+        iconStyle: {
+            transition: "all .3s ease-in 0s",
+            color: "#007DB8",
+        },
+    };
+    handleClick = () => {
+        const {toggleContent} = this.state;
+        this.setState({toggleContent: !toggleContent});
```
alexander-rassanov: it is not safety to upgrade so. Since the setState will add this changes to event loop, and before how it will be executed the state can be already changed.
setState can accept a function where the first argument is state. So we can get
`this.setState(({toggleContent})=> ({
    toggleContent: !toggleContent,
)})`

src/portlet/Portlet.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+export class Portlet extends React.PureComponent<PortletProps, PortletState> {
+    state: PortletState = {
+        toggleContent: true,
+        iconStyle: {
+            transition: "all .3s ease-in 0s",
+            color: "#007DB8",
+        },
+    };
+    handleClick = () => {
```
alexander-rassanov: missed blank line above

src/portlet/Portlet.tsx
```diff
@@ -0,0 +1,94 @@
...
+        const {toggleContent} = this.state;
+        this.setState({toggleContent: !toggleContent});
+        toggleContent
```
alexander-rassanov: I think it is better to use if, since you don't need the result of this.setState and you just process the logic

src/portlet/Portlet.tsx
vinodwani18: addressed

src/portlet/Portlet.tsx
vinodwani18: addressed

src/portlet/Portlet.tsx
vinodwani18: addressed

src/portlet/Portlet.tsx
```diff
@@ -0,0 +1,97 @@
...
+                  iconStyle: {
+                      transition: "all .3s ease-in 0s",
+                      float: "right",
+                      color: "#007DB8",
+                  },
+              });
+    };
+
+    render() {
+        const {className, style, header, children} = this.props;
```
psivakumaran: Where are we using 'style' prop? Is it applicable to the children or the main'div'?

src/portlet/Portlet.tsx
```diff
@@ -0,0 +1,97 @@
...
+        const {toggleContent, iconStyle} = this.state;
+        return (
+            <div
+                style={{
```
psivakumaran: I feel we should move the styles to a .css file?

src/portlet/Portlet.tsx
vinodwani18: Right now we are not using it but we may need it when we will implement this component in Vsphere as we dont have vmware styling here.

src/portlet/Portlet.tsx
vinodwani18: Addressed -created  new file classnames.tsx and exported styles object with required styles

src/portlet/Portlet.tsx
vinodwani18: Addressed - removed style props

src/portlet/ClassNames.tsx
```diff
@@ -0,0 +1,34 @@
...
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+export class Styles {
+    public static ICON_STYLE = {transition: "all .3s ease-in 0s", color: "#007DB8"};
+    public static MAIN_DIV: any = {
```
akate1: No need to add explicit 'any' type here.

src/portlet/Portlet.stories.tsx
```diff
@@ -0,0 +1,57 @@
...
+import * as React from "react";
+import {storiesOf} from "@storybook/react";
+import {Portlet} from ".";
+
+storiesOf("Portlet", module).add("Basic Portlet", () => (
+    <div className="clr-row">
+        <div className="clr-col-lg-4 clr-col-md-8 clr-col-12">
+            <Portlet header="Header- Porlet Component">
+                <div>Portlet Component </div>
+                <table className="table">
```
akate1: Can we use our 'Table' component here ?

src/portlet/Portlet.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+import * as React from "react";
+import {classNames} from "../utils";
+import {Styles} from "./ClassNames";
+import {Icon, Direction} from "../icon";
+
+type PortletProps = {
+    style?: any;
+    className?: string;
+    header: any;
```
akate1: In story we are passing string to header prop then why we need any type here ?

src/portlet/ClassNames.tsx
vinodwani18: yes, it is required at it is throwing error for some of the properties

src/portlet/Portlet.tsx
vinodwani18: addressed

src/portlet/Portlet.stories.tsx
vinodwani18: addressed

PR #40, opened 2019-08-28
Flex 144 port datagrid component to react

psivakumaran:
Changes for porting data grid component to clarity-repo
1) Basic data grid structure
2) Data grid with custom elements
3) Data grid with multi-select
4) Updated check box component to support multi-select option in data grid
5) Added data grid with batch action
6) Fix issue of rendering radio button without label
7) fix issue of hide/show of Button component

### DIFF
--- a/src/datagrid/DataGrid.stories.tsx
+++ b/src/datagrid/DataGrid.stories.tsx
@@ -40,6 +40,16 @@ storiesOf("DataGrid", module)
             />
         </div>
     ))
+    .add("Grid with single select option", () => (
+        <div style={{width: "80%"}}>
+            <DataGrid
+                columns={normalColumns}
+                data={normalRows}
+                footer={footer}
+                selectionType={GridSelectionType.SINGLE}
+            />
+        </div>
+    ))
     .add("Grid with batch action", () => (
         <div style={{width: "80%"}}>
             <GridActions ref={datagridActionsRef} />
--- a/src/datagrid/DataGrid.stories.tsx
+++ b/src/datagrid/DataGrid.stories.tsx
@@ -10,42 +10,38 @@
 
 import * as React from "react";
 import {storiesOf} from "@storybook/react";
-import {
-    DataGridColumn,
-    DataGridRow,
-    DataGridCell,
-    DataGridFooter,
-    DataGrid,
-    DataGridColumnHeader,
-    DataGridRowContainer,
-    DataGridBody,
-} from "./DataGrid";
-storiesOf("DataGrid", module).add("Basic", () => (
-    <div style={{width: "80%"}}>
-        <DataGrid>
-            <DataGridBody>
-                <DataGridColumnHeader>
-                    <DataGridColumn>{"User ID"}</DataGridColumn>
-                    <DataGridColumn>{"Name"}</DataGridColumn>
-                    <DataGridColumn> {"Creation date"}</DataGridColumn>
-                    <DataGridColumn>{"Favorite color"}</DataGridColumn>
-                </DataGridColumnHeader>
-                <DataGridRowContainer>
-                    <DataGridRow>
-                        <DataGridCell> {41512} </DataGridCell>
-                        <DataGridCell> {"Georgia"} </DataGridCell>
-                        <DataGridCell> {"Sep 11, 2018"} </DataGridCell>
-                        <DataGridCell> {"Blue"} </DataGridCell>
-                    </DataGridRow>
-                    <DataGridRow>
-                        <DataGridCell> {9524} </DataGridCell>
-                        <DataGridCell> {"Brynn"} </DataGridCell>
-                        <DataGridCell> {"Aug 2, 2014"} </DataGridCell>
-                        <DataGridCell> {"Purple"} </DataGridCell>
-                    </DataGridRow>
-                </DataGridRowContainer>
-            </DataGridBody>
-            <DataGridFooter> {"10 users"} </DataGridFooter>
-        </DataGrid>
-    </div>
-));
+import {Icon} from "../icon";
+import {DataGrid, GridSelectionType} from "./DataGrid";
+import {normalColumns, normalRows, customRows, footer} from "./DataGridValues";
+
+storiesOf("DataGrid", module)
+    .add("Basic grid", () => (
+        <div style={{width: "80%"}}>
+            <DataGrid columns={normalColumns} data={normalRows} footer={footer} />
+        </div>
+    ))
+    .add("Grid with custom cells", () => (
+        <div style={{width: "80%"}}>
+            <DataGrid columns={normalColumns} data={customRows} footer={footer} />
+        </div>
+    ))
+    .add("Grid with single select option", () => (
+        <div style={{width: "80%"}}>
+            <DataGrid
+                columns={normalColumns}
+                data={normalRows}
+                footer={footer}
+                selectionType={GridSelectionType.SINGLE}
+            />
+        </div>
+    ))
+    .add("Grid with multi select option", () => (
+        <div style={{width: "80%"}}>
+            <DataGrid
+                columns={normalColumns}
+                data={normalRows}
+                footer={footer}
+                selectionType={GridSelectionType.MULTI}
+            />
+        </div>
+    ));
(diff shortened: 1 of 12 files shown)


src/datagrid/DataGrid.tsx
```diff
@@ -0,0 +1,400 @@
...
+        allColumns: this.props.columns,
+        allRows: this.props.data,
+    };
+
+    /* ##########  DataGrid public methods start  ############ */
+    // Function to return all selected rows
+    getSelectedRows(): DataGridRow[] {
+        const {allRows} = this.state;
+        let selectedRows = new Array();
+        if (this.state.selectAll) selectedRows = allRows;
```
alexander-rassanov: let's add brackets and have
if () {
} else {

src/datagrid/DataGrid.tsx
```diff
@@ -0,0 +1,400 @@
...
+        else {
+            allRows.forEach(row => {
```
alexander-rassanov: It can be transformed to more functional/immutable way
```
selectedRows = allRows.filter(({isSeleted}) => isSelected);
```

src/datagrid/DataGrid.tsx
```diff
@@ -0,0 +1,400 @@
...
+    };
+
+    // Initial state of datagrid
+    state: DataGridState = {
+        selectAll: false,
+        allColumns: this.props.columns,
+        allRows: this.props.data,
+    };
+
+    /* ##########  DataGrid public methods start  ############ */
```
alexander-rassanov: I think the function doesn't have private keyword and it is enough to understand it is a public method

src/datagrid/DataGrid.tsx
```diff
@@ -0,0 +1,400 @@
...
+        if (this.state.selectAll) selectedRows = allRows;
+        else {
+            allRows.forEach(row => {
+                if (row["isSelected"]) selectedRows.push(row);
+            });
+        }
+        return selectedRows;
+    }
+
+    /* ##########  DataGrid public methods end  ############ */
```
alexander-rassanov: same. exceeded comment

alexander-rassanov: probably is to have just a convention
For example
- class Name
- statis fields
- state
- local fields
- constructor
- public methods
- private methods

src/utils.ts
```diff
@@ -14,7 +14,18 @@ export function classNames(classNameList: (false | undefined | null | string)[])
...
 
+//Find if all the keys in an object are true
 export function allTrue(obj: any) {
     for (var key in obj) if (!obj[key]) return false;
     return true;
 }
+
+// Find if a particular key in an object array is true for all elements in the array.
+export function allTrueOnKey(obj: any, key: any) {
+    for (var index in obj) {
```
alexander-rassanov: use let and `in` can be replaced on `of` to iterate through values

src/datagrid/DataGrid.tsx
akate1: done.

src/datagrid/DataGrid.tsx

src/datagrid/DataGrid.tsx

src/utils.ts
akate1: done

src/datagrid/DataGrid.stories.tsx
```diff
@@ -0,0 +1,68 @@
...
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+import * as React from "react";
+import {storiesOf} from "@storybook/react";
+import {Icon} from "../icon";
```
ben-schumacher: unused import

src/datagrid/DataGrid.stories.tsx
```diff
@@ -0,0 +1,68 @@
...
+import {Button} from "../forms/button";
```
ben-schumacher: unused import

src/datagrid/DataGrid.tsx
```diff
@@ -0,0 +1,396 @@
...
+    }
+
+    /* ##########  DataGrid private methods start  ############ */
+    // Function to handle select/deselect of all rows
+    private handleSelectAll = (evt: React.ChangeEvent<HTMLInputElement>) => {
+        const rows = this.state.allRows;
+        const value = this.state.selectAll;
+        const {onSelectAll} = this.props;
+        rows.forEach(row => (row["isSelected"] = !value));
+        this.setState({
```
ben-schumacher: I think because updating state is async, we should be more careful and raise `onSelectAll` in callback?  Like so ```this.setState(
            {
                selectAll: !value,
                allRows: rows,
            },
            () => {
                onSelectAll && onSelectAll();
            },
        );```

src/datagrid/DataGrid.tsx
```diff
@@ -0,0 +1,396 @@
...
+        const {onRowSelect, selectionType} = this.props;
+        rows.forEach(row => {
+            if (row["rowID"] === rowID) {
+                const value = !row["isSelected"];
+                row["isSelected"] = value;
+            } else if (selectionType === GridSelectionType.SINGLE) {
+                row["isSelected"] = false;
+            }
+        });
+        this.setState({
```
ben-schumacher: Maybe use this instead: ```this.setState(
            {
                allRows: [...rows],
                selectAll: allTrueOnKey(rows, "isSelected"),
            },
            () => {
                onRowSelect && onRowSelect();
            },
        );```

src/datagrid/DataGrid.stories.tsx
```diff
@@ -0,0 +1,68 @@
...
+    .add("Grid with batch action", () => (
+        <div style={{width: "80%"}}>
+            <GridActions ref={datagridActionsRef} />
+            <DataGrid
+                ref={datagridRef}
+                columns={normalColumns}
+                data={normalRows}
+                footer={footer}
+                selectionType={GridSelectionType.MULTI}
+                onRowSelect={() => {
```
ben-schumacher: If we don't also add onSelectAll here, the delete button shows incorrect count when I check/uncheck select all check box.  I think we always need to be updating GridActions state.

src/datagrid/DataGrid.tsx
```diff
@@ -0,0 +1,396 @@
...
+                </div>
+            </div>
+        );
+    }
+
+    // Function to build datagrid header
+    private buildDataGridHeader(): React.ReactElement {
+        const {columns, selectionType} = this.props;
+        return (
+            <div className={ClassNames.DATAGRID_HEADER} role="rowgroup">
```
ben-schumacher: Checking VMWare Clarity website, this div should get an additional class `datagrid-selected` when the checkbox is selected.

src/datagrid/DataGridValues.tsx
```diff
@@ -0,0 +1,183 @@
...
+            showEdit: rows.length === 1 ? true : false,
+        });
+    }
+
+    render() {
+        const {selectedRows, showEdit} = this.state;
+        return (
+            <div>
+                <Button key="new">NEW</Button>
+                {
```
ben-schumacher: can remove bracket here

src/datagrid/DataGrid.stories.tsx
akate1: done

src/datagrid/DataGrid.stories.tsx
akate1: done

src/datagrid/DataGrid.tsx

src/datagrid/DataGrid.stories.tsx
akate1: Added custom function for SelectAll.

src/datagrid/DataGrid.tsx
akate1: @ben-schumacher  Thanks for noticing this. the 'datagrid-selected' gets added to row when row is selected.

src/datagrid/DataGridValues.tsx
akate1: done.

src/datagrid/DataGrid.tsx

PR #45, opened 2019-09-20
- Added functionality of hide and show columns with datagrid

akate1:
- Updated version of @clr/ui to 2.0.0 to get latest CSS classes for datagrid
- Added functionality of hide and show columns with datagrid

image

### DIFF
--- a/src/datagrid/DataGrid.stories.tsx
+++ b/src/datagrid/DataGrid.stories.tsx
@@ -24,7 +24,7 @@ import {
     paginationDetails,
     paginationRows,
     pageFilterFunction,
-    hideNshowCol,
+    hideableColumns,
 } from "./DataGridValues";
 import {CustomFilter} from "./CustomFilter";
 
@@ -237,7 +237,7 @@ storiesOf("DataGrid", module)
     .add("Grid with HIde and Show Column", () => (
         <div style={{width: "80%", paddingTop: "5%"}}>
             <DataGrid
-                columns={hideNshowCol}
+                columns={hideableColumns}
                 rows={normalRows}
                 footer={{
                     footerData: "Total 2 users",
--- a/src/datagrid/DataGrid.stories.tsx
+++ b/src/datagrid/DataGrid.stories.tsx
@@ -24,6 +24,7 @@ import {
     paginationDetails,
     paginationRows,
     pageFilterFunction,
+    hideNshowCol,
 } from "./DataGridValues";
 import {CustomFilter} from "./CustomFilter";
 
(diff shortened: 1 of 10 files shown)


src/datagrid/HideShowColumns.tsx
```diff
@@ -0,0 +1,239 @@
...
+                            style={{
+                                top: 0,
+                                bottom: "auto",
+                                right: "auto",
+                                left: 0,
+                                transform: transformVal,
+                            }}
+                        >
+                            <div className={ClassNames.SWITCH_HEADER}>
+                                Select Columns
```
payald: Can the text "Select Columns" be part of en.json ?

src/datagrid/HideShowColumns.tsx
```diff
@@ -0,0 +1,239 @@
...
+                            </ul>
+
+                            <div className={ClassNames.SWITCH_FOOTER}>
+                                <Button
+                                    className={ClassNames.SWITCH_BUTTON}
+                                    link
+                                    disabled={SelectAll}
+                                    onClick={this.handleSelectAll}
+                                >
+                                    Select All
```
payald: Can the text "Select All" be part of en.json ?

src/datagrid/HideShowColumns.tsx
```diff
@@ -0,0 +1,239 @@
...
+            if (el && typeof el !== "string" && !el.contains(target)) {
+                this.toggle();
+            }
+        }
+    };
+
+    // Handle for select All button
+    private handleSelectAll = () => {
+        this.updateDatagridColumns("All");
+        this.toggle();
```
payald: Query: On clicking SELECT ALL, the menu disappears. Although in clarity design the menu remains open on clicking SELECT ALL. Do we need this.toggle() after updateDatagridColumns() within handleSelectAll ?

src/datagrid/HideShowColumns.tsx
```diff
@@ -0,0 +1,239 @@
...
+    };
+
+    // Handle for select single column
+    private handleSingleSelect = (columnName: string) => {
+        this.updateDatagridColumns(columnName);
```
payald: Query: In clarity design when only one checkbox remains checked then the last checkbox cannot be unchecked. In this case, the last checkbox also gets unchecked. Can we add a condition such that at least one checkbox remains checked so that at least one column is always visible ?

src/datagrid/DataGrid.stories.tsx
```diff
@@ -233,15 +234,27 @@ storiesOf("DataGrid", module)
             />
         </div>
     ))
+    .add("Grid with HIde and Show Column", () => (
```
payald: Observation: On navigating to the second page and then selecting all the check boxes, when we again navigate to the first page then only the select all checkbox is selected while the other checkboxes are not selected.

src/datagrid/DataGrid.stories.tsx
```diff
@@ -24,6 +24,7 @@ import {
     paginationDetails,
     paginationRows,
     pageFilterFunction,
+    hideNshowCol,
```
payald: Can we rename hideNshowCol and/or hideShowCol to hideableColumn in all the references ?

src/datagrid/HideShowColumns.tsx
akate1: done

src/datagrid/DataGrid.stories.tsx
akate1: done. thanks for noticing this.

src/datagrid/DataGrid.stories.tsx
akate1: done

PR #47, opened 2019-09-24
Feature flex 822 fix wizard for vsphere

ben-schumacher:
My proposed solution to include custom components in the footer is to use `children`.  Please let me know your thoughts.  Sample below to render the wizard within vSphere.
`<Wizard
                size={WizardSize.MEDIUM}
                show={true}
                showNav={false}
                isInline={true}
                title={osintl.intl.formatMessage({id: "rules.new.title"})}
                steps={SingleStep}
                onClose={() => this.handleClose()}
                onFinish={() => this.handleApply()}
                closable={false}
                finishButtonText="apply"
                cancelButtonText="cancel"
                style={{height: "600px"}}
            >
                <CheckBox label="Create another" />
            </Wizard>`

### DIFF
--- a/src/datagrid/DataGrid.tsx
+++ b/src/datagrid/DataGrid.tsx
@@ -15,6 +15,7 @@ import {CheckBox} from "../forms/checkbox";
 import {RadioButton} from "../forms/radio";
 import {Button} from "../forms/button";
 import {Icon, Direction} from "../icon";
+import {HideShowColumns} from "./HideShowColumns";
 
 /**
  * General component description :
@@ -33,14 +34,15 @@ import {Icon, Direction} from "../icon";
  * @param {footer} footer component
  * @param {onRowSelect} Function which will gets called on select/deselect of rows
  * @param {onSelectAll} Function which will gets called on select/deselect of all rows
+ * @param {keyfield} field to uniquely identify row
  * @param {rowType} Expandable or compact row type
  * @param {itemText} label to display for all items
+ * @param {pagination} pagination support
  */
 type DataGridProps = {
     className?: string;
     style?: any;
     selectionType?: GridSelectionType;
-    pagination?: boolean;
     columns: DataGridColumn[];
     rows?: DataGridRow[];
     footer?: DataGridFooter;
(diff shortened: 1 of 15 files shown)


ben-schumacher:
NewRule

src/datagrid/DataGrid.tsx
```diff
@@ -953,7 +951,7 @@ export class DataGrid extends React.PureComponent<DataGridProps, DataGridState>
             >
                 {footer && footer.hideShowColBtn && this.buildHideShowColumnsBtn()}
                 <div className={ClassNames.DATAGRID_FOOTER_DESC}>
-                {pagination && pagination ? this.buildDataGridPagination(): this.buildFooterContent() }
+                    {pagination && pagination ? this.buildDataGridPagination() : this.buildFooterContent()}
```
alexander-rassanov: double pagination check

src/datagrid/DataGrid.tsx
```diff
@@ -927,19 +927,17 @@ export class DataGrid extends React.PureComponent<DataGridProps, DataGridState>
...
+    private buildFooterContent(): React.ReactElement {
         const {footer} = this.props;
         const {allRows, itemText} = this.state;
         const footerDescription = allRows.length.toString() + " " + itemText;
         let content;
 
-        if (footer !== undefined){
-            content =  footer.footerData !== undefined ?  footer.footerData : footerDescription;
+        if (footer !== undefined) {
+            content = footer.footerData !== undefined ? footer.footerData : footerDescription;
```
alexander-rassanov: Do you expect footerData can have 0,'' (empty)values and is it correct?
If not we can replace that one on `content = footer.footerData || footerDescription`

alexander-rassanov:
LGTM. left some comments

src/datagrid/DataGrid.tsx
akate1: @alexander-rassanov 'footerData' is an optional prop so it can be undefined.

src/datagrid/DataGrid.tsx
akate1: @alexander-rassanov  thanks for noticing this. @ben-schumacher  can you pleas take care of this DataGrid code as well ?

PR #50, opened 2019-09-26
Updated single step story of wizard to show use of onStepSubmit prop

akate1:
image

### DIFF
--- a/src/wizards/CustomStepComponent.tsx
+++ b/src/wizards/CustomStepComponent.tsx
@@ -1,0 +1,40 @@
+/**
+ * Copyright (c) 2018 Dell Inc., or its subsidiaries. All Rights Reserved.
+ *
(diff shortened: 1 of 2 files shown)


src/wizards/CustomStepComponent.tsx
```diff
@@ -0,0 +1,40 @@
...
+export class CustomStepComponent extends React.PureComponent<any, CustomStepComponentState> {
+    state: CustomStepComponentState = {
+        value: " ",
+    };
+
+    resetComponent() {
+        this.setState({value: " "});
+    }
+
+    private handleChnage = (evt: React.ChangeEvent<any>) => {
```
psivakumaran: handleChnage => handleChange

akate1: done. Thanks for noticing this Priya

PR #51, opened 2019-10-04
Fixed issue of datagrid not getting updated after mutation

akate1:


ben-schumacher:
Just curious what mutation needed this?

akate1:
> Just curious what mutation needed this?
@ben-schumacher  After we fire event acknowledgement mutation. The Relay store gets updated and it called render() of EventList component with updated data.
In this scenario we are getting updated event list through props  but as datagrid was not aware of any props change it was showing old data. To fix this we are using 'componentWillReceiveProps()' in datagrid so that it will get updated props and change its state accordingly.

src/datagrid/DataGrid.tsx
```diff
@@ -243,6 +243,16 @@ export class DataGrid extends React.PureComponent<DataGridProps, DataGridState>
         if (this.props.pagination !== undefined) this.setInitalStateForPagination();
     }
 
+    componentWillReceiveProps(nextProps: DataGridProps) {
```
vinodwani18: componentWillReceiveProps is marked as depricated, So i think here we can use "getDerivedStateFromProps"instead of using componentWillReceiveProps.

akate1: @vinodwani18  We know that and I have given a try for 'getDerivedStateFromProps' as well. But 'getDerivedStateFromProps' is a static method and we can not call instance methods from static. Even after making instance methods static we got some weird results for sorting and filtering. 
We have already discuss this internally and decided to go with 'componentWillReceiveProps' for now.
Along with 'componentWillReceiveProps' , 'componentWillMount' will also going to deprecate in future.  So we will log an enhancement bug for all common components to change deprecated methods.

alexander-rassanov: use componentDidUpdate

src/datagrid/DataGrid.tsx
```diff
@@ -516,18 +521,27 @@ export class DataGrid extends React.PureComponent<DataGridProps, DataGridState>
 
     private updateRowIDs(rows: DataGridRow[]) {
         // set rowID = index in array
-        rows.map((row: DataGridRow, index: number) => {
+        rows.forEach(function(row: DataGridRow, index: number) {
```
alexander-rassanov: this function is not pure. Use map instead. Here you modify the object that is a reason of changes of the incoming arg

src/datagrid/DataGrid.tsx
```diff
@@ -516,18 +521,27 @@ export class DataGrid extends React.PureComponent<DataGridProps, DataGridState>
...
+
         return rows;
     }
 
-    private updateColumnIDs(cols: DataGridColumn[]) {
+    private updateColumnIDs(columns: DataGridColumn[]) {
         // set columnID = index in array
-        cols.map((cols: DataGridColumn, index: number) => {
-            cols["columnID"] = index;
+        columns.forEach(function(column: DataGridColumn, index: number) {
```
alexander-rassanov: same as above -> map

src/datagrid/DataGrid.tsx
akate1: @alexander-rassanov  'componentDidUpdate' life cycle hook gets execute after render() and we need some changes before that. Still I tried with it but it is no showing desired behavior. With this method the data grid is getting disappeared on any click.

src/datagrid/DataGrid.tsx

src/datagrid/DataGrid.tsx

PR #52, opened 2019-10-07
Fixed issue of finish button with sync single step wizard

akate1:
- Issue -  finish button was not getting disabled in case of Wizard with single step and validation type is sychronous validation.
- Modified story for same
- Remvoe console.log from VerticalNav.tsx

image

### DIFF
--- a/src/wizards/CustomStepComponent.tsx
+++ b/src/wizards/CustomStepComponent.tsx
@@ -10,12 +10,18 @@
 
 import * as React from "react";
 import {Input} from "../forms/input/Input";
+import {Button} from "../forms/button";
 
 type CustomStepComponentState = {
     value: string;
 };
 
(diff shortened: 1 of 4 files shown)


src/wizards/CustomStepComponent.tsx
```diff
@@ -31,9 +37,20 @@ export class CustomStepComponent extends React.PureComponent<any, CustomStepComp
 
     render() {
         const {value} = this.state;
+        const {wizardRef, stepId} = this.props;
         return (
             <div style={{width: "166px"}}>
+                {"Enter something"} :
```
alexander-rassanov: Why you added brackets here?

akate1: removed brackets

PR #53, opened 2019-10-10
Adding id and name for Input

girishdudhe:
name filed in missing in Input component and id is not correctly assigned.So corrected the code to add id and name in Input.

### DIFF
--- a/src/forms/input/Input.tsx
+++ b/src/forms/input/Input.tsx
@@ -10,6 +10,7 @@
 
 import * as React from "react";
 import * as utils from "../../utils";
+import {UID} from "react-uid";
 import {ReactNode} from "react";
 
 type InputProps = {
@@ -22,6 +23,7 @@ type InputProps = {
     onChange?: (evt: React.ChangeEvent<HTMLInputElement>) => void;
     placeholder?: string;
     name: string;
+    id?: string;
     value?: string;
     defaultValue?: string;
     size?: number;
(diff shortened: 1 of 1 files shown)


src/forms/input/Input.tsx
```diff
@@ -61,46 +63,58 @@ export class Input extends React.PureComponent<InputProps> {
...
+                            id={setId ? setId : id}
+                            name={name}
                             value={value}
                             defaultValue={defaultValue}
                             size={size}
                             disabled={disabled}
-                            className="clr-input"
                             placeholder={placeholder}
                             onChange={this.handleChange}
+                            className={className}
```
akate1: I think we also need to keep class name "clr-input" and then add user define class names after that.

src/forms/input/Input.tsx
```diff
@@ -61,46 +63,58 @@ export class Input extends React.PureComponent<InputProps> {
...
+                            <div className="clr-input-wrapper">
+                                <input
+                                    type={type || "text"}
+                                    name={name}
+                                    id={setId ? setId : id}
+                                    value={value}
+                                    defaultValue={defaultValue}
+                                    size={size}
+                                    disabled={disabled}
+                                    className="clr-input"
```
akate1: Same here keep class name 'clr-input' and append user defined class names after that

src/forms/input/Input.tsx
girishdudhe: I have not changes .... so dropping the comment

src/forms/input/Input.tsx
girishdudhe: dropping comment as per above

PR #54, opened 2019-10-11
Add onClick functionality for toggle

psivakumaran:


src/forms/toggle/Toggle.tsx
```diff
@@ -21,6 +21,7 @@ type ToggleProps = {
     checked?: boolean;
     defaultChecked?: boolean;
     onChange?: (newValue: boolean) => void;
+    onClick?: (evt: React.MouseEvent<HTMLInputElement>) => void;
```
akate1: Since it is toggle component I guess onChange is sufficient. is there any specific case where you need onClick ?

alexander-rassanov: Please, add default props for undefined function. It can be just () => {}

src/forms/toggle/Toggle.tsx
```diff
@@ -49,7 +50,7 @@ export class Toggle extends React.PureComponent<ToggleProps, ToggleState> {
         }
     }
     render() {
-        const {label, name, ariaLabelledby, disabled} = this.props;
+        const {label, name, ariaLabelledby, disabled, onClick} = this.props;
```
alexander-rassanov: place the destruction to different lines, please.

psivakumaran: I did, but before the commit, the linter re-arranges it

src/forms/toggle/Toggle.tsx
psivakumaran: Done

src/forms/toggle/Toggle.tsx
psivakumaran: I did, but the linter re-arranges it before the commit

PR #57, opened 2019-10-21
Added width and error props to Input

akate1:
This PR has taken care of following things:
- Modify Input component to accept custom styles and error handling
- created stories in Input for same
- optimized Input code to avoid code duplication
-  Added name and required props in select

image

image

### DIFF
--- a/src/forms/input/Input.tsx
+++ b/src/forms/input/Input.tsx
@@ -76,7 +76,7 @@ export class Input extends React.PureComponent<InputProps> {
                 <input
                     type={type || "text"}
                     name={name}
-                    id={id ? id : uid}
+                    id={id || uid}
                     value={value}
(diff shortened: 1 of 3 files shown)


src/forms/input/Input.tsx
```diff
@@ -50,43 +56,67 @@ export class Input extends React.PureComponent<InputProps> {
...
             id,
+            required,
+            onBlur,
+        } = this.props;
+        return (
+            <React.Fragment>
+                <input
+                    type={type || "text"}
+                    name={name}
+                    id={id ? id : uid}
```
alexander-rassanov: you can replace it by >
```suggestion
                    id={id || uid}
```
to go through js conventions

akate1: done

PR #58, opened 2019-10-22
Flex 1661 port password component to react

akate1:
image

image

### DIFF
--- a/src/forms/password/Password.tsx
+++ b/src/forms/password/Password.tsx
@@ -33,21 +33,24 @@ type PasswordProps = {
     minPasswordLength?: number;
     required?: boolean; // auto-check on blur if there's a value
     error?: boolean; // force error state of component
-    showPassword?: boolean;
+    unmask?: boolean; // if true renders eye icon to hide/show or mask/unmask password
 };
 
 type PasswordState = {
     show: boolean;
     type: string;
+    value: any;
 };
+
 export class Password extends React.PureComponent<PasswordProps, PasswordState> {
     static defaultProps = {
-        showPassword: true,
+        unmask: true,
     };
 
     state: PasswordState = {
         show: false,
         type: "password",
+        value: null,
     };
 
     private showHidePassword = () => {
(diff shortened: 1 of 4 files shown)


src/forms/password/Password.tsx
```diff
@@ -0,0 +1,141 @@
...
+                    {label && Password.renderLabel(label)}
+                    <div className={utils.classNames(classNames)} style={{width: "100%"}}>
+                        <div className="clr-input-wrapper">
+                            <div className="clr-input-group clr-row" style={{...style, paddingRight: "1rem"}}>
+                                <input
+                                    maxLength={maxPasswordLength}
+                                    minLength={minPasswordLength}
+                                    name={name}
+                                    defaultValue={defaultValue}
+                                    value={value}
```
psivakumaran: When we enter the value in the password component, and if I do not have a onChange which sets the value prop explicitly, how will I be able to get the password value? I think value should be part of the password state

src/forms/password/Password.tsx
```diff
@@ -0,0 +1,141 @@
...
+                <div className={utils.classNames(["clr-form-control", label && "clr-row"])}>
+                    {label && Password.renderLabel(label)}
+                    <div className={utils.classNames(classNames)} style={{width: "100%"}}>
+                        <div className="clr-input-wrapper">
+                            <div className="clr-input-group clr-row" style={{...style, paddingRight: "1rem"}}>
+                                <input
+                                    maxLength={maxPasswordLength}
+                                    minLength={minPasswordLength}
+                                    name={name}
+                                    defaultValue={defaultValue}
```
psivakumaran: Generally we should not have a default value for password

src/forms/password/Password.tsx
akate1: done

src/forms/password/Password.tsx
akate1: keeping this for now. There may be scenario where we get password from back end and user need to change it.

psivakumaran: Agreed

PR #59, opened 2019-10-23
Changed title type to any to support component use

psivakumaran:


psivakumaran:
> Please remove private from navigationClick() we need this method in Review page of new ObjectStore.

Done

PR #60, opened 2019-11-04
Made totalItems parameter of updateRows() optional  for pagination

akate1:
For pagination the totalItems are total items present in back-end and not just length of data-gird rows.
So changed logic in updateRows() accordingly.

### DIFF
--- a/src/datagrid/DataGrid.tsx
+++ b/src/datagrid/DataGrid.tsx
@@ -246,7 +246,7 @@ export class DataGrid extends React.PureComponent<DataGridProps, DataGridState> 
     componentDidUpdate(prevProps: DataGridProps) {
         const {rows, columns} = this.props;
         if (rows && rows !== prevProps.rows) {
-            this.updateRows(rows, rows.length);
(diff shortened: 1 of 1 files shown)


src/datagrid/DataGrid.tsx
```diff
@@ -267,12 +267,12 @@ export class DataGrid extends React.PureComponent<DataGridProps, DataGridState>
...
 
     // Function to update datagrid rows
-    updateRows = (rows: DataGridRow[], totalItems: number) => {
+    updateRows = (rows: DataGridRow[], totalItems?: number) => {
         const updatedRows = this.updateRowIDs(rows);
         let {pagination} = this.state;
 
         // update pagination footer
-        if (pagination) {
+        if (pagination && totalItems) {
```
psivakumaran: In which case the totalItems will be different from rows.length()?

akate1: @psivakumaran I have mention that in PR description. For pagination total-items are different than length of current rows.

PR #67, opened 2019-11-27
Feature flex 1666 data qa attribute

girishdudhe:
Added DataQA field and Props in Components SignPost, Modal, Wizard, Toast and Layout

### DIFF
--- a/src/wizards/Wizard.tsx
+++ b/src/wizards/Wizard.tsx
@@ -83,6 +83,13 @@ export enum WizardValidationType {
     NONE = "None",
 }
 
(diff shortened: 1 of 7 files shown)


src/wizards/Wizard.tsx
```diff
@@ -544,6 +572,15 @@ export class Wizard extends React.PureComponent<WizardProps> {
         );
     }
 
+    updateDataQAStrings(dataqa: string) {
+        let prefix = "prefix";
```
akate1: let -> const

PR #75, opened 2020-02-10
Fixed issue of select error icon

akate1:
**Description:**
FLEX-2161 - Common component - Select field - Error icon colliding with the dropdown arrow

**Bug:**
https://asdjira.isus.emc.com:8443/browse/FLEX-2161

Storybook:
image

Vsphere Plugin:

image

### DIFF
--- a/src/forms/select/Select.tsx
+++ b/src/forms/select/Select.tsx
@@ -12,6 +12,7 @@ import * as React from "react";
 import {UID} from "react-uid";
 import {Icon} from "../../icon";
 import {classNames} from "../../utils";
+import "./Select.css";
 
 type SelectOption = {
     value?: string;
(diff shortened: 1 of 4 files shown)


src/forms/select/Select.tsx
```diff
@@ -72,14 +71,28 @@ export class Select extends React.PureComponent<SelectProps> {
...
-                style={{width: error ? "75%" : "100%"}}
+                style={{width: this.getSelectWidth()}}
             >
                 {!showDefaultSelect && <option selected disabled hidden style={{display: "none"}} value="" />}
                 {children}
             </select>
         );
     }
 
+    // Function to calulate width of select tag
```
alexander-rassanov: What do you think if we would have 2 classes with 95% and 100% and we could just return a class to use it?
Why? - it will be more simple to find several parameters of width in one place(above/below of the component)

akate1: @alexander-rassanov I don't mind creating CSS classes for same but this repo uses CSS of Vmware Clarity and we have not created our custom classes so far.

maheshkulkarni4: @akate1  - I think, we should create custom classes here. We should not use custom styling unless really necessary.  So below method can be updated to return a class instead of direct style.

akate1: done

akate1: done

PR #81, opened 2020-02-26
Added new props for navigation click on wizard left pane links

vinodwani18:
image

### DIFF
--- a/src/wizards/Wizard.tsx
+++ b/src/wizards/Wizard.tsx
@@ -53,6 +53,7 @@ type WizardStepNavDetails = {
     stepNavIcon?: string;
     stepNavTitle?: string;
     stepNavChildren?: React.ReactNode;
+    onNavClick?: () => Promise<any>;
 };
 
 /**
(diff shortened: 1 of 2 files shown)


src/wizards/Wizard.tsx
```diff
@@ -407,6 +408,8 @@ export class Wizard extends React.PureComponent<WizardProps> {
     }
 
     navigationClick(stepId: number) {
+        const {onNavClick} = this.props;
+        onNavClick && onNavClick(stepId);
```
akate1: This will call onNavClick() asynchronously. Are we  ok with it ? Also please add a story for same.

src/wizards/Wizard.tsx
```diff
@@ -101,6 +101,7 @@ type WizardProps = {
     previousButtonClassName?: string;
     nextButtonText?: string;
     onNext?: () => Promise<any>;
+    onNavClick?: Function;
```
akate1: This function should be part of WizardNavLinkDetails type. As each NavLink may have different callback function

src/wizards/Wizard.tsx
vinodwani18: addressed

src/wizards/Wizard.tsx
vinodwani18: addressed

PR #85, opened 2020-03-11
Added additional  props in button and dropdwon component

akate1:
- Added 'value' prop in Button and Dropdown-Item component
- Added 'active' prop in Dropdown-Item to identify active item
- Added direction to open menu in Dropdown

image

### DIFF
--- a/src/forms/dropdown/Dropdown.tsx
+++ b/src/forms/dropdown/Dropdown.tsx
@@ -20,25 +20,39 @@ import {STOP_PROPAGATION} from ".";
 
 /**
  * DropDown Props
- * @param {label} label of checkbox
+ * @param {label} label of dropdown
  * @param {isNested} property to nest dropdown
  * @param {className} css property
  * @param {button} dropdown button
  * @param {showCaret} caret property
+ * @param {direction} direction to open a menu
  * @param {dataqa} quality engineering testing field
  */
 export type DropdownProps = {
-    label?: string;
+    label?: any;
     isNested?: boolean;
     className?: string;
     button?: ButtonProps;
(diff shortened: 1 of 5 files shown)


src/forms/dropdown/Dropdown.tsx
```diff
@@ -20,26 +20,40 @@ import {STOP_PROPAGATION} from ".";
 
 /**
  * DropDown Props
- * @param {label} label of checkbox
+ * @param {label} label of dropdown
  * @param {isNested} property to nest dropdown
  * @param {className} css property
  * @param {button} dropdown button
  * @param {showCaret} caret property
+ * @param {direction} direction to open a menu
```
psivakumaran: Small suggestion - More than 'direction to open menu' it should be 'position of the dropdown'

src/forms/dropdown/DropdownItem.tsx
```diff
@@ -16,10 +16,12 @@ import {Button} from "../button";
 export type DropdownItemProps = {
     menuItemType: MenuItemType;
     label?: string;
+    value?: string;
     key?: string;
     isDisabled?: boolean;
     isExpandable?: boolean;
     isHeaderChild?: boolean;
+    active?: boolean;
```
psivakumaran: Should this be named as 'isActive'?

akate1: done

src/forms/dropdown/Dropdown.tsx
akate1: We have used same naming convention in tooltip and signpost so keeping this as it is.

PR #87, opened 2020-03-12
Properly import Moment

lpetkov:
use  import * as Moment from 'moment' instead of Moment from "moment"

Also: make the Signpost border visible (by using the proper class name)

And: add babel-preset-react-app as a dependency, since Storybook returned an error message after running yarn upgrade. See the error message here: https://pastebin.com/FCTWPt4s


maheshkulkarni4:
This PR is open for long.  We havent seen any issues in import so far.

PR #131, opened 2020-08-25
[OBSDEF-1308] datagrid column width

akate1:
**Description:**
- Added feature in datagrid to change width of datagrid column using width dragger

**Bug:**
- https://jira.cec.lab.emc.com/browse/OBSDEF-1308

**Screen Shots:**
Original Grid
image

After increasing column width:
image

After decreasing column width:
image

Min width indicator: 
image

### DIFF
--- a/src/datagrid/DataGrid.tsx
+++ b/src/datagrid/DataGrid.tsx
@@ -908,7 +908,7 @@ export class DataGrid extends React.PureComponent<DataGridProps, DataGridState> 
         return (
             <div
                 role="columnheader"
-                className={classNames([ClassNames.DATAGRID_COLUMN, className])}
+                className={classNames([ClassNames.DATAGRID_COLUMN, ClassNames.DATAGRID_NG_STAR_INSERTED, className])}
                 aria-sort="none"
                 style={{...style, width: width + "px"}}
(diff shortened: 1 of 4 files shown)


src/datagrid/DataGrid.tsx
```diff
@@ -884,12 +902,15 @@ export class DataGrid extends React.PureComponent<DataGridProps, DataGridState>
     // Function to build datagrid colums
     private buildDataGridColumn(column: DataGridColumn, index: number): React.ReactElement {
         const {columnName, columnID, className, style, sort, filter, width} = column;
+        const columnHeight =
```
maheshkulkarni4, line 905: If the refs are null by any chance, this will be set to null.  Do we want to set some default height instead of null ?

maheshkulkarni4 (approved):
LGTM

src/datagrid/DataGrid.tsx
akate1: @maheshkulkarni4  This ref will never be null even if there is no data in datagrid. In worst case if it is null the height will set as "auto". Please refer line no 167 at DataGridColumnResize.tsx

girishdudhe (approved):

PR #134, opened 2020-09-14
[OBSDEF-2806] Tab component enhancements

akate1:
**Description:**
- Enhancement to tab component to support static tab and tab name with icon in it
- Add provision of callback function on tab click
- Updated stories in storybook
- Fixed CSS issues

**JIRA:**
- https://jira.cec.lab.emc.com/browse/OBSDEF-2806

**Testing:** 
 - Vertical Tabs
image

- Horizontal Tabs with single disable tab
image

- With management as static tab
image

- With overflow tabs
image


girishdudhe (approved):

maheshkulkarni4 (approved):
Code looks good.  

One scenario came up in discussion which we need to handle as enhancement. 
If we select a tab which is inside the Dropdown, then user cant see the selected tab.  So possible option is to show the selected tab always on the Tabs bar and move some other component into the dropdown.  This is more like a dynamic behaviour.  But we want to have it.  @girishdudhe lets add a enhancement story to cover this use case.

akate1:
> Code looks good.
> 
> One scenario came up in discussion which we need to handle as enhancement.
> If we select a tab which is inside the Dropdown, then user cant see the selected tab. So possible option is to show the selected tab always on the Tabs bar and move some other component into the dropdown. This is more like a dynamic behaviour. But we want to have it. @girishdudhe lets add a enhancement story to cover this use case.

@girishdudhe and @maheshkulkarni4 I have taken care of that scenario. You can see working of it in 'overflow tabs' story. As per clarity the overflow tab gets selected for any active tab in drop-down.  If we want to change this clarity behavior then we can add enhancement story for it

Clarity Document : https://clarity.design/documentation/tabs

PR #137, opened 2020-09-23
[obsdef-3016] -  Added custom size for modals and added storybook

maheshkulkarni4:
JIRA: https://jira.cec.lab.emc.com/browse/OBSDEF-3016

Feature:  Model component was supporting fix sizes for the modal sizing.  Due to this the wrong sizes were getting applied to the Modal Dialogs in case of Standalone application.  So we have added Custom size, where the width and height are directly applied to the Modal. 

**Screenshots:**
with custom style
image

With Custom Classname
image

### DIFF
--- a/src/modals/Modal.tsx
+++ b/src/modals/Modal.tsx
@@ -24,6 +24,7 @@ import {ReactNode} from "react";
  * @param {dataqa} Quality Engineering field
  * @param {width} if Size is custom, then width need to be provided in props
  * @param {height} if Size is custom, then height need to be provided in props
+ * @param {className} if className is provided, the add custom class with existing classes
  */
 type ModalProps = {
     isOpen?: boolean;
@@ -34,6 +35,7 @@ type ModalProps = {
     dataqa?: string;
     width?: number;
     height?: number;
+    className?: string;
 };
 
 type ModalState = {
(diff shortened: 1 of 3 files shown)


girishdudhe (approved):

girishdudhe (approved):

src/modals/Modal.tsx
```diff
@@ -30,6 +32,8 @@ type ModalProps = {
     closable?: boolean;
     onClose?: Function;
     dataqa?: string;
+    width?: number;
```
akate1, line 35: @maheshkulkarni4 along with this can you please add `className `and style `props `? Those will be useful in future as well.

src/modals/Modal.tsx
```diff
@@ -100,6 +105,7 @@ export class Modal extends React.PureComponent<ModalProps> {
                             ClassNames.MODAL_DIALOG, //prettier
                             size && size,
```
akate1, line 106: we can use our 'className' prop here to achieve same result

src/modals/Modal.tsx
```diff
@@ -100,6 +105,7 @@ export class Modal extends React.PureComponent<ModalProps> {
...
                         ])}
+                        style={size === ModalSize.CUSTOM ? {width: `${width}px`, height: `${height}px`} : {}}
```
akate1, line 108: Please add option to use specified styles as well

akate1 (approved):
@maheshkulkarni4 Please update version in package.json 

LGTM, Have given some minor comments

src/modals/Modal.tsx
maheshkulkarni4: We dont want user to send this custom style from components, so not adding support for style props.  Discussed with @akate1  already.

src/modals/Modal.tsx
maheshkulkarni4: Added a new Prop for ClassName.  As discussed with @akate1 , we are not adding prop for custom inline styling. Let developers use the custom classes to achieve same.

PR #140, opened 2020-10-06
[OBSDEF-2071] datagrid expandable rows enhancement

akate1:
**Description:**
- Add `isLoading ` prop for datagrid
- Provided callback function and lazy loading for row expansion

**JIRA:**
- https://jira.cec.lab.emc.com/browse/OBSDEF-2071

**Testing:**
image

image

image

### DIFF
--- a/src/datagrid/DataGrid.tsx
+++ b/src/datagrid/DataGrid.tsx
@@ -102,9 +102,24 @@ export type DataGridRow = {
     style?: any;
     rowID?: number; // not to take from user
     isSelected?: boolean;
+    expandableRowData?: expandableRowDetails;
+    disableRowSelection?: boolean;
+};
+
+/**
+ * type for datagrid expandable row data :
+ * @param {isLoading} if true then show loading icon for expandable row
+ * @param {onClick} callback function to  fetch expandable row contents
+ * @param {expandableContent} content to show after row expand
+ * @param {isExpanded} true if row is already expanded
+ * @param {hideRowExpandIcon} if true then hide icon for expandable row
+ */
+export type expandableRowDetails = {
+    isLoading?: boolean;
+    onClick?: (row: DataGridRow) => Promise<any>;
     expandableContent?: any;
     isExpanded?: boolean;
-    disableRowSelection?: boolean;
+    hideRowExpandIcon?: boolean;
 };
 
 /**
@@ -131,7 +146,7 @@ export type DataGridFooter = {
     footerData?: any;
     className?: string;
     style?: any;
-    hideShowColumns?: DataGRidHideShowColumns;
+    hideShowColumns?: DataGridHideShowColumns;
     showFooter: boolean;
 };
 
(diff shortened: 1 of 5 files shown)


src/datagrid/DataGrid.tsx
```diff
@@ -43,6 +43,7 @@ import {DataGridColumnResize} from "./DataGridColumnResize";
  * @param {selectedRowCount} number of selected rows across all pages
  * @param {dataqa} quality engineering tag
  * @param {id} unique ID for datagrid
+ * @param {isLoading} if true shows loading spinner else shows datagrid
```
maheshkulkarni4, line 46: Comment can be updated, its only showing loadingSpinner on a single row, that too at the CheckBox's place. 

I have some questions 

1) if one roo is in expanded state,  backend call is going on and if user expands another row then do we keep both rows in expanded state.  (We should I think)

2) If the expanded row makes backend call, which takes longer time to return.  where do we show error messgaes and how the Expanded rows behave in this case.

src/datagrid/DataGridValues.tsx
```diff
@@ -126,15 +126,27 @@ export const customRows = [
 const expandableContent =
     "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Proin in neque in ante placerat mattis id sed quam. Proin rhoncus lacus et tempor dignissim. Vivamus sem quam, pellentesque aliquet suscipit eget, pellentesque sed arcu. Vivamus in dui lectus. Suspendisse cursus est ac nisl imperdiet viverra. Aenean sagittis nibh lacus, in eleifend urna ultrices et. Mauris porttitor nisi nec velit pharetra porttitor. Vestibulum vulputate sollicitudin dolor ut tincidunt. Phasellus vitae blandit felis. Nullam posuere ipsum tincidunt velit pellentesque rhoncus. Morbi faucibus ut ipsum at malesuada. Nam vestibulum felis sit amet metus finibus hendrerit. Fusce faucibus odio eget ex vulputate rhoncus. Fusce nec aliquam leo, at suscipit diam.";
 
-export const expandableRows = [
+//Custom function to filter data
+export const loadExpandableContent = (rows: DataGridRow): Promise<any> => {
+    return new Promise((resolve, reject) => {
+        // Purposefully added dealy here to see loading spinner
```
maheshkulkarni4, line 132: dealy = delay

maheshkulkarni4 (approved):
LGTM

src/datagrid/DataGridValues.tsx
akate1: done

src/datagrid/DataGrid.tsx
akate1: @maheshkulkarni4  
**For Question 1 :** If user expands more than one row we are keeping all rows in expanded state until user collapse it. This behaviour is as per VMware clarity.

**For Question 2 :** In case of error Parent component who is sending callback should send error to datagrid and datagrid can show that error in expandable row section. Vmware clarity do not have any such use case as of now. But we can support it.

girishdudhe (approved):
LGTM

PR #144, opened 2020-10-19
[OBSDEF-3306] [OBSDEF-3410] Added loading spinner in wizard step and fix console errors

akate1:
**Description :** 
 - Added loading spinner in wizard step
- Fixed below console errors on datagrid and select component

console errors: 
`Warning: Can't call setState on a component that is not yet mounted. This is a no-op, but it might indicate a bug in your application. Instead, assign to `this.state` directly or define a `state = {};` class property with the desired state in the DataGrid component.
    in DataGrid (at DataGrid.stories.tsx:88)
    in div (at DataGrid.stories.tsx:87)
    in div (created by Story)`

`react-dom.development.js:88 Warning: Use the `defaultValue` or `value` props on <select> instead of setting `selected` on <option>.
    in option (at DataGrid.tsx:1147)
    in select (at DataGrid.tsx:1140)`

**JIRA:**
- https://jira.cec.lab.emc.com/browse/OBSDEF-3306
- https://jira.cec.lab.emc.com/browse/OBSDEF-3410

**Testing:** 
image

image

image

### DIFF
--- a/src/datagrid/DataGrid.tsx
+++ b/src/datagrid/DataGrid.tsx
@@ -293,12 +293,12 @@ export class DataGrid extends React.PureComponent<DataGridProps, DataGridState> 
         }
     }
 
-    // funtion to initialize datagrid state
+    // Function to initialize datagrid state
     initializeDataGridState = (): DataGridState => {
         const {isLoading, itemText} = this.props;
         const rows = this.initializeRowData();
         const columns = this.initializeColumnData();
-        const state: DataGridState = {
+        const dataGridState: DataGridState = {
             isLoading: isLoading || false,
             selectAll: this.isAllRowsSelected(rows),
             allColumns: [...columns],
@@ -306,7 +306,7 @@ export class DataGrid extends React.PureComponent<DataGridProps, DataGridState> 
             itemText: itemText || DEFAULT_ITEM_TEXT,
             pagination: this.initializePaginationData(),
         };
-        return state;
+        return dataGridState;
     };
 
     // Function to initialize rows data in state
--- a/src/datagrid/DataGrid.tsx
+++ b/src/datagrid/DataGrid.tsx
@@ -228,6 +228,18 @@ export enum GridRowType {
 }
 
 const isSelectedKey = "isSelected";
+
+// Default label for datagrid items
+const DEFAULT_ITEM_TEXT: string = "items";
+
+// Default width of datagrid column in px
+export const DEFAULT_COLUMN_WIDTH: number = 100;
+
+// Default pagination constants
+export const DEFAULT_CURRENT_PAGE_NUMBER: number = 1;
+export const DEFAULT_PAGE_SIZE: number = 10;
+export const DEFAULT_TOTAL_ITEMS: number = 0;
+
 /**
  * State for DataGrid :
  * @param {selectAll} set to true if all rows got selected else false
@@ -257,9 +269,6 @@ type DataGridPaginationState = {
     compactFooter?: boolean;
 };
 
-// Default width of datagrid column in px
-export const DEFAULT_COLUMN_WIDTH = 100;
-
 /**
  * DataGrid Componnet :
  * Displays data in grid format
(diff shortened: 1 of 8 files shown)


src/datagrid/DataGrid.tsx
```diff
@@ -228,6 +228,18 @@ export enum GridRowType {
 }
 
 const isSelectedKey = "isSelected";
+
+// Default label for datagrid items
+const DEFAULT_ITEM_TEXT: string = "items";
```
maheshkulkarni4, line 233: Why do we need defauly text ?  isnt this something which should come as Prop just like other labels ?

src/forms/select/Select.tsx
```diff
@@ -14,7 +14,7 @@ import {Icon} from "../../icon";
 import {classNames} from "../../utils";
 
 type SelectOption = {
-    value?: string;
+    value?: any;
```
maheshkulkarni4, line 17: Why are we switching from String to any ?  Any opens door for more problems as strict checking is close to disabled in this case.

src/newWizard/WizardStep.tsx
```diff
@@ -52,6 +53,7 @@ export default class WizardStep extends React.PureComponent<WizardStepProps> {
                 aria-hidden={id !== currentStepID}
                 className={classNameList}
             >
+                {isLoading && <Spinner size={SpinnerSize.MEDIUM} />}
```
maheshkulkarni4, line 56: Just thinking, do we need the spinnerSize to be a Prop instead ?

src/datagrid/DataGrid.tsx
```diff
@@ -312,6 +293,76 @@ export class DataGrid extends React.PureComponent<DataGridProps, DataGridState>
...
+
+            const firstItem: number = this.getFirstItemIndex(currentPageNumber, datagridPageSize);
+            const lastItem: number = this.getLastItemIndex(datagridPageSize, totalItemsInDatagrid, firstItem);
+
+            const paginationState: DataGridPaginationState = {
+                currentPage: currentPageNumber,
+                pageSize: datagridPageSize,
+                totalItems: totalItemsInDatagrid,
+                pageSizes: pageSizes,
+                compactFooter: compactFooter || false,
```
maheshkulkarni4, line 356: Might be old code, but why is this called CompactFooter and why not Footer ?
do we have Footer(big) as well ?

src/datagrid/DataGrid.tsx
```diff
@@ -312,6 +293,76 @@ export class DataGrid extends React.PureComponent<DataGridProps, DataGridState>
...
+    };
+
+    // Function to initialize rows data in state
+    initializeRowData = (): DataGridRow[] => {
+        const {rows} = this.props;
+        let updatedRows: DataGridRow[] = [];
+        if (rows && rows.length) {
+            updatedRows = this.updateRowIDs(rows);
+            updatedRows.forEach(function(row) {
+                const rowSelectionIsDisabled = row.disableRowSelection !== undefined ? row.disableRowSelection : false;
```
maheshkulkarni4, line 319: **row.disableRowSelection !== undefined  ? ** 
can be replaced with 
**row.disableRowSelection ?**  as well.

maheshkulkarni4 (approved):
LGTM

src/datagrid/DataGrid.tsx
akate1: @maheshkulkarni4 This is for compact footer which we show when objectStoreDetails page is open

image

src/datagrid/DataGrid.tsx
akate1: @maheshkulkarni4 `disableRowSelection `is a `Boolean `field So here I wanted to check if user has passed it or not.

src/forms/select/Select.tsx
akate1: @maheshkulkarni4  `Value` prop has `any` type in Select component and here it was string. So I changed it to any. With this we can create Select of type number , string or any other options. Let me know your thoughts on this.

src/datagrid/DataGrid.tsx
akate1: @maheshkulkarni4 This it default text that will get displayed in footer if user has not send any label to display. 
For example, 10 items vs 10 users. Should I make it a mandatory prop ?

src/newWizard/WizardStep.tsx
akate1: done

src/datagrid/DataGrid.tsx
```diff
@@ -312,6 +293,76 @@ export class DataGrid extends React.PureComponent<DataGridProps, DataGridState>
         }
     }
 
+    // funtion to initialize datagrid state
```
payald, line 296: Typo funtion -> function

src/datagrid/DataGrid.tsx
```diff
@@ -312,6 +293,76 @@ export class DataGrid extends React.PureComponent<DataGridProps, DataGridState>
...
+    initializeDataGridState = (): DataGridState => {
+        const {isLoading, itemText} = this.props;
+        const rows = this.initializeRowData();
+        const columns = this.initializeColumnData();
+        const state: DataGridState = {
```
payald, line 301: Can we rename state to dataGridState?

src/datagrid/DataGrid.tsx

payald (approved):

PR #155, opened 2020-12-21
Fixed issue of custom filters in datagrid

akate1:
**Description:**
- Added support for multi select custom filter in datagrid
-  Fixed issue of not able to preserve filtered value in custom filter

**JIRA:**
- https://jira.cec.lab.emc.com/browse/OBSDEF-4529

**Testing:**
customFilterIssue

### DIFF
--- a/src/datagrid/CustomFilterMulti.tsx
+++ b/src/datagrid/CustomFilterMulti.tsx
@@ -1,0 +1,76 @@
+/**
+ * Copyright (c) 2020 Dell Inc., or its subsidiaries. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+import * as React from "react";
+import {CheckBox} from "../forms/checkbox";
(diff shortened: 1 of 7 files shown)


src/datagrid/CustomFilterMulti.tsx
```diff
@@ -0,0 +1,76 @@
...
+            },
+            onChange && onChange(values),
+        );
+    };
+
+    render() {
+        const {values} = this.state;
+
+        return (
+            <div style={{width: "130px", padding: "1em"}}>
```
maheshkulkarni4, line 62: Is this test code ?  labels are hard coded.  Check if thi need to be moved to stories.

maheshkulkarni4, line 62: Convert inline styles to classes

maheshkulkarni4 (approved):
LGTM

src/datagrid/CustomFilterMulti.tsx
akate1: @maheshkulkarni4 Can't do this in clarity-react for now. Also, this code is just for story/demo so keeping it as it is

akate1: @maheshkulkarni4 Can't do this in clarity-react for now. Also, this code is just for story/demo so keeping it as it is

PR #165, opened 2021-02-08
Feature - Introduced Debounce Behaviour for Input Field and DataList 

maheshkulkarni4:
**Problem Statement : **

- OnChange, onKeydown, onKeyPress handlers on **Input** and **DataList** trigger as soon as event occurs.  This was causing problems when we use Input, Datalist in the filters.  
- Each keypress, keydown or change event triggers the Backend calls which may be bad for the Functionalities like Events, Issues where each backend call is a costly operations as too many backend calls may make the backend service unstable.
-
**Solution Proposed: **
- Added debounce behaviour for Input and DataList.  
- Default Debounce Time is 500ms.  So the event will only get sent out only after user idle time of 500ms
- This will reduce the backend calls to great extent.  

**Input Field - Stories with Debounce: **
Debounce-with-input-fields

**DataList - Stories with Debounce: **
Debounce-with-datalist


src/forms/common/DebounceUtils.tsx
```diff
@@ -0,0 +1,38 @@
...
+     * @param evt - SyntheticEvent Event which is input to the called function
+     * @param func - Function to be applied post debouce time is over
+     * @param debounceTime - optional debounceTime in miliseconds, defaultDebounce will be used if not provided
+     */
+    public debounce = (evt: React.SyntheticEvent, func: Function | undefined, debounceTime: number | undefined) => {
+        // if function passed is not set by parent
+        if (!func) {
+            return func;
+        }
+        const waitTime = debounceTime || defaultDebounce;
```
payald, line 27: If debounceTime=0 is passed then defaultDebounce=500 will be set. Is this by design ?

payald (approved):

src/forms/common/DebounceUtils.tsx
maheshkulkarni4: if debounce 0 is passed, it will use 0 and setTimeout triggers immediately when i get chance. 
500 is applied only when the debounceTime is not set at all.  Its safe to not set debounceTime at all and let component set 500ms by default.  

User need to set this when they want longer or lesser debounce time then default.

src/forms/common/DebounceUtils.tsx
```diff
@@ -0,0 +1,38 @@
+/**
+ * Copyright (c) 2018 Dell Inc., or its subsidiaries. All Rights Reserved.
```
akate1, line 2: 2018 -> 2021

src/forms/common/DebounceUtils.tsx
```diff
@@ -0,0 +1,38 @@
...
+ *
```
akate1, line 3: can we change extension of this file from tsx -> ts ??

akate1 (approved):
LGTM

girishdudhe (approved):
LGTM

src/forms/common/DebounceUtils.tsx
maheshkulkarni4: Done

src/forms/common/DebounceUtils.tsx
maheshkulkarni4: Done

PR #167, opened 2021-02-11
Feature [OBSDEF-5812] - debounce behaviour enhacements

maheshkulkarni4:
In 0.25.4 build, the debounce was made mandatory for all text fields. This had sideeffect on the Input with Type="number".  So in this PR, we have introduced one more Prop for **DataList** and **Input** components.
- debounce   ->  optional field, boolean. This need to be set to false if you dont want debounce on the input field
- debounceTime -> optional field, number. Default debounce Time is 500ms, if you want to change that time, then you can set these numbers. 

**Debounce Set False**
Debounce-set-off-input-field-with-numbers

**Debounce Set True - with 1 Second debounceTime ** 
Debounce-set-on-input-field-numbers

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@dell/clarity-react",
-    "version": "0.25.4",
+    "version": "0.25.5",
     "description": "React components for Clarity UI",
     "license": "Apache-2.0",
     "private": false,
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@dell/clarity-react",
-    "version": "0.25.5",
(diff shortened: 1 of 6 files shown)


ghatea (approved):
LGTM.

package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@dell/clarity-react",
-    "version": "0.25.4",
+    "version": "0.25.5",
```
payald, line 3: 0.25.5->0.25.6

payald (approved):

girishdudhe (approved):
LGTM

PR #169, opened 2021-02-25
Added debounce feature to Filters and also added fix for clearing filter

maheshkulkarni4:
https://jira.cec.lab.emc.com/browse/OBSDEF-4746

1) Added Debounce of 500ms

Filters on ObjectStoreList, events, issues pages fires too many backend calls, as every keystroke fires one backend call. This is not really performant, so to reduce this behaviour we have added debounce with 500ms to all the text based filters.

2) DataGridFilter didnt had option to clear themself, so exposed a method to clear the dataGrid Filter by passing Empty string

3) Introducted **disabled** prop to allow disabling filter textbox if the props is applied. This is important to avoid user typing into the textbox when some backend calls are already in progress and loader is visible.  In current implementation, the DataGrid Loader does not block user from keep typing into filter.  So this disabled prop would help solve this issue.

### DIFF
--- a/src/datagrid/DataGridFilter.tsx
+++ b/src/datagrid/DataGridFilter.tsx
@@ -13,6 +13,7 @@ import {classNames} from "../utils";
 import {ClassNames} from "./ClassNames";
 import {Button} from "../forms/button";
 import {DataGridRow} from "./DataGrid";
+import {DebounceUtils} from "../forms/common/DebounceUtils";
 
 /**
  * Props for DataGridFilter :
@@ -23,6 +24,7 @@ import {DataGridRow} from "./DataGrid";
  * @param {placeholder} placeholder for string filter input
  * @param {onFilter} Custom filter logic
  * @param {filterType} Type of filter string or custom
+ * @param {disabled} boolean value to enable or disable filter
  */
 export type DataGridFilterProps = {
     style?: any;
@@ -33,6 +35,7 @@ export type DataGridFilterProps = {
     onFilter: (rows: DataGridRow[], columnValue: any, columnName: string) => Promise<DataGridFilterResult>;
     filterType?: FilterType;
     showFilter?: boolean;
+    disabled?: boolean;
 };
 
 /**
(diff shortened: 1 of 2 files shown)


src/datagrid/DataGridFilter.tsx
```diff
@@ -251,9 +264,9 @@ export class DataGridFilter extends React.PureComponent<DataGridFilterProps, Dat
     }
 
     render() {
-        const {isOpen} = this.state;
+        const {isOpen, isFiltered} = this.state;
         const {showFilter} = this.props;
-        const {isFiltered} = this;
+        //const {isFiltered} = this;
```
girishdudhe, line 269: remove commented line

girishdudhe (approved):
LGTM ...

src/datagrid/DataGridFilter.tsx
```diff
@@ -237,9 +249,10 @@ export class DataGridFilter extends React.PureComponent<DataGridFilterProps, Dat
                             name={`name-${columnName}`}
                             placeholder={placeholder}
                             defaultValue={filterValue}
-                            onChange={evt => {
-                                this.updateFilter(evt.target.value);
-                            }}
+                            disabled={disabled || false}
+                            onChange={(evt: React.ChangeEvent<HTMLInputElement>) =>
```
akate1, line 253: can we add a prop to enable/disable `debounce `effect ?

akate1 (approved):
LGTM

src/datagrid/DataGridFilter.tsx
```diff
@@ -129,6 +136,11 @@ export class DataGridFilter extends React.PureComponent<DataGridFilterProps, Dat
         }
     }
 
+    private handlOnChange = (evt: React.ChangeEvent<HTMLInputElement>) => {
```
psivakumaran, line 139: Minor comment 
handlChange -> handleChange

psivakumaran (approved):
LGTM.

src/datagrid/DataGridFilter.tsx
maheshkulkarni4: Added now. I dont expect to users to use it.  Default 500ms is good for standard usecases. But good to have to have option if we want to disable it.

src/datagrid/DataGridFilter.tsx
maheshkulkarni4: Addressed.

src/datagrid/DataGridFilter.tsx
maheshkulkarni4: Addressed

PR #185, opened 2021-06-01
Added functionality for option group in select component

ghatea:
Description: 

Added otpion to have option group in Select component

Screenshot:
select-option-group

### DIFF
--- a/src/forms/select/Select.stories.tsx
+++ b/src/forms/select/Select.stories.tsx
@@ -115,7 +115,7 @@ storiesOf("Select", module)
     ))
     .add("Select Option Group", () => (
         <div>
-            <Select onChange={evt => console.log("hh", evt.target.value)}>
+            <Select onChange={action("Option group select - change")}>
                 <SelectOptionGroup label="Group 1">
                     <SelectOption value="1">One</SelectOption>
                     <SelectOption value="2">Two</SelectOption>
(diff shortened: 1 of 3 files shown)


src/forms/select/Select.stories.tsx
```diff
@@ -112,4 +112,20 @@ storiesOf("Select", module)
                 <SelectOption value="3">Three</SelectOption>
             </Select>
         </div>
+    ))
+    .add("Select Option Group", () => (
+        <div>
+            <Select onChange={evt => console.log("hh", evt.target.value)}>
```
akate1, line 118: hh -> Selected Option

src/forms/select/Select.tsx
```diff
@@ -28,6 +28,19 @@ export const SelectOption: React.FunctionComponent<SelectOption> = ({value, sele
     );
 };
 
+type SelectOptionGroup = {
+    label: string;
+    disabled?: boolean;
+};
```
akate1, line 34: Please add documentation for this

src/forms/select/Select.tsx
```diff
@@ -28,6 +28,19 @@ export const SelectOption: React.FunctionComponent<SelectOption> = ({value, sele
...
+
+export const SelectOptionGroup: React.FunctionComponent<SelectOptionGroup> = ({label, disabled, children}) => {
+    return (
+        <optgroup label={label} disabled={disabled}>
+            {children}
+        </optgroup>
```
akate1, line 40: if disable is true here. Will it disable child options as well ?
Can u add story for disabled group option ?

ghatea: Yes It will disabled child options as well. Added the story as well.
disabled-option-group

src/forms/select/Select.stories.tsx
ghatea: Removed console. Added action instead.

src/forms/select/Select.tsx
ghatea: Done.

akate1 (approved):

maheshkulkarni4 (approved):
LGTM

PR #205, opened 2021-09-09
Adds onBlur event on Password Component

rnmulchandani:
`onBlur` event was missing on the `Password` Component.
Added the `onBlur` event which will be passed to the `Password` component as a prop.

### DIFF
--- a/src/forms/password/Password.tsx
+++ b/src/forms/password/Password.tsx
@@ -89,6 +89,7 @@ export class Password extends React.PureComponent<PasswordProps, PasswordState> 
             label,
             value,
             defaultValue,
+            onBlur,
             onKeyPress,
             title,
             errorTitle,
@@ -130,6 +131,7 @@ export class Password extends React.PureComponent<PasswordProps, PasswordState> 
                                     required={required}
                                     onKeyPress={onKeyPress}
                                     title={title}
+                                    onBlur={onBlur}
                                     type={type}
                                     disabled={disabled}
                                     style={{width: "95%"}}
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@dellstorage/clarity-react",
-    "version": "0.25.28",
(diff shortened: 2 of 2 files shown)


src/forms/password/Password.tsx
```diff
@@ -130,6 +131,7 @@ export class Password extends React.PureComponent<PasswordProps, PasswordState>
                                     required={required}
                                     onKeyPress={onKeyPress}
                                     title={title}
+                                    onBlur={onBlur}
```
maheshkulkarni4, line 134: Password component should already have a stoybook.  Add a Test case this.

onKeyPress -> triggers the change immediately after user types anything.  Why is onBlur needed here ? 
One problem with onBlur -> 

User types password, but never press tab -> Directly clicks Login buttin (in case of login screen) -> The onBlur wont fire.  Then what exact usecase we are solving here ? 

 if you set the Callback handler for onKeyPress, you keep password in your state and when user change it, you always have updated copy.  Please check if we really need onBlur ?

maheshkulkarni4 (changes requested):
Please check the comments.

src/forms/password/Password.tsx
rnmulchandani: The `onBlur` event is needed when we are using a `Password` component in the form.
Let's say the user clicked on the Password component but didn't provide any value for it, once the user leaves the component we can show an error as "Password cannot be blank" based on the `touched` property.

```
<Password
             onChange={props.handleChange}
             onBlur={props.handleBlur}
             name="password"
           />
```
Ref: https://formik.org/docs/api/formik

The test case is already available for this in the storybook.
https://github.com/EMCECS/clarity-react/blob/master/src/forms/password/Password.stories.tsx#L53

akate1 (approved):

maheshkulkarni4 (approved):
LGTM

PR #219, opened 2021-12-03
Update version to 1.0.0

bhardg21:
create release branch and update version to 1.0.0


maheshkulkarni4 (changes requested):
I think, we dont need this PR.
The branch `release-objectscale-1.0` need to live along side `Master`

PR #220, opened 2021-12-03
add scripts for release branch

bhardg21:
Add script to execute job for release branch


maheshkulkarni4 (approved):
I think, we should not update existing Jobs.  We need new files which would create Jobs.  I prefer adding new files on the MASTER and then adding same on this branch too.

maheshkulkarni4 (changes requested):
Approved by mistake earliar.

bhardg21:
> I think, we should not update existing Jobs. We need new files which would create Jobs. I prefer adding new files on the MASTER and then adding same on this branch too.

I will create new file, let's have discussion for the same

PR #235, opened 2022-03-16
[OBSDEF-16020] Correct direction of expand and collapse vertical nav group

akate1:
The `caret ` icon direction should be downward for expansion and right for the collapse. But for the vertical nav group, it was showing it incorrectly. 

image


psivakumaran (approved):

tejasjc-ecs (approved):
LGTM

PR #242, opened 2022-05-05
[obsdef-17570] - Added ComponentDidUpdate method to force update the tabs if any change happen in props

maheshkulkarni4:
[obsdef-17570] - Added ComponentDidUpdate method to force update the tabs if any change happen in props


tejasjc-ecs (approved):
LGTM

psivakumaran (approved):

PR #244, opened 2022-05-31
[OBSDEF-15071] fix data grid filter position

Bhatii:
## Summary
This PR fixes:
- Detached filter from header issue
- Set default filter position to left aligned
- Console warning issues
- Using URLs attribute in html issue

### After fix:

https://user-images.githubusercontent.com/78294852/171174193-24f2ca98-7a61-423f-a1e5-b8a5604a9fd6.mp4


src/datagrid/DataGrid.tsx
```diff
@@ -1131,7 +1131,7 @@ export class DataGrid extends React.PureComponent<DataGridProps, DataGridState>
         const showEmptyHeader: boolean = rowType && rowType !== GridRowType.COMPACT ? true : false;
 
         return (
-            <div className={ClassNames.DATAGRID_HEADER} role="rowgroup">
+            <div className={ClassNames.DATAGRID_HEADER} style={{position: "relative"}} role="rowgroup">
```
GauravMenezes-Professional, line 1134: Can we put this styling in a CSS class?

GauravMenezes-Professional (approved):
LGTM, please check comment

src/constants.ts
```diff
@@ -0,0 +1,19 @@
...
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+/**
+ * Constant for Icon center position offset in pixels
+ */
+export const ICON_CENTER_POSITION = 12;
```
akate1, line 14: If these constants are related to datagird  filter then please move them in their respective file

src/datagrid/CustomFilterMulti.tsx
```diff
@@ -63,15 +63,14 @@ export class CustomFilterMulti extends React.PureComponent<DataGridCustomFilterP
...
-                    return (
-                        <CheckBox
-                            label={color}
-                            onClick={evt => this.handleFilterChange(evt, color)}
-                            checked={values.includes(color)}
-                        />
-                    );
-                })}
+                {COLOR_OPTIONS.map((color: string, index: number) => (
+                    <CheckBox
```
akate1, line 67: don't we need to return from map()?

src/datagrid/DataGridFilter.tsx
```diff
@@ -8,27 +8,28 @@
...
+ * @param {any} datagridRef - Reference for DataGrid on which filter will gets apply. We need this to call method which will update datagird rows.
+ * @param {string} columnName - columnName on which filter will apply
+ * @param {string} placeholder - placeholder for string filter input
+ * @param {Function} onFilter - Custom filter logic
+ * @param {FilterType} filterType - Type of filter string or custom
+ * @param {boolean} disabled - boolean value to enable or disable filter
+ * @param {boolean} debounce - boolean value to apply debounce behavior
+ * @param {number} debounceTime - number value debounceTime/Delay value in milliseconds
+ * @param {FilterPosition} position - position of the filter popup
+ * @param {any} defaultValue - defaultValue of the filter
```
akate1, line 32: can u please re-check all comment changes? its showing datatype first and then prop name

src/utils.ts
```diff
@@ -39,3 +42,25 @@ export function allFalseOnKey(obj: any, key: any) {
...
 }
+
+/**
+ * Get X axis position value for supplied object
+ * @param {sting} position - Position prop, Accepted values right, left, center
+ * @param {number} childWidth - Width of child element in pixels, for center and left alignment
+ * @returns {number} x axis position
+ * */
+
+export const getXPositionValue = (position: string, childWidth: number): number => {
```
akate1, line 53: If we are using this function only for datagrid filter then please move it into that file

src/utils.ts
```diff
@@ -39,3 +42,25 @@ export function allFalseOnKey(obj: any, key: any) {
...
+
+    if (position === FilterPosition.RIGHT) {
+        return ICON_CENTER_POSITION;
+    }
+    if (position === FilterPosition.CENTER) {
+        return ICON_CENTER_POSITION - childWidth / 2;
+    }
+    if (position === FilterPosition.LEFT) {
+        return ICON_CENTER_POSITION - childWidth;
+    }
```
akate1, line 64: you can replace  these if statements with switch case

akate1 (changes requested):

src/datagrid/DataGridFilter.tsx
```diff
@@ -8,27 +8,29 @@
...
+ * @param {*} datagridRef - Reference for DataGrid on which filter will gets apply. We need this to call method which will update datagird rows.
+ * @param {string} columnName - columnName on which filter will apply
+ * @param {string} placeholder - placeholder for string filter input
+ * @param {function} onFilter - Custom filter logic
+ * @param {FilterType} filterType - Type of filter string or custom
+ * @param {boolean} disabled - boolean value to enable or disable filter
+ * @param {boolean} debounce - boolean value to apply debounce behavior
+ * @param {number} debounceTime - number value debounceTime/Delay value in milliseconds
+ * @param {FilterPosition} position - position of the filter popup
+ * @param {*} defaultValue - defaultValue of the filter
```
akate1, line 33: Please revert these comments related changes

akate1 (changes requested):

src/datagrid/DataGridUtils.ts
```diff
@@ -0,0 +1,32 @@
...
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+import {ICON_HEIGHT} from "./constants";
+import {FilterPosition} from "./DataGridFilter";
+
+/**
+ * Get X axis position value for supplied object
+ * @param {sting} position - Position prop, Accepted values right, left, center
```
maheshkulkarni4, line 16: If the accepted values are fixed, then introduce a enum and use that enum instead of string as type for this prop

src/datagrid/DataGridUtils.ts
```diff
@@ -0,0 +1,32 @@
...
+ * @param {number} childWidth - Width of child element in pixels, for center and left alignment
+ * @returns {number} X axis position relative to parent element
+ * */
+
+export const getXPositionValue = (position: string, childWidth: number): number => {
+    if (!position) return 0;
+
+    const {CENTER, LEFT, RIGHT} = FilterPosition;
+    return position === RIGHT
```
maheshkulkarni4, line 25: Change this to a switch case if possible. Would look better than if/else hell.

src/datagrid/constants.ts
```diff
@@ -0,0 +1,19 @@
+/**
```
maheshkulkarni4, line 1: Follow the nameing pattern.  All files under this directory has prefix DataGrid... use same for this too.

src/datagrid/DataGridUtils.ts
maheshkulkarni4, line 25: Check and add story book to cover these scenarios if already not present.

src/utils.ts
```diff
@@ -8,7 +8,7 @@
  *     http://www.apache.org/licenses/LICENSE-2.0
  */
 
-export type ReactChildren = (React.ReactNode) | (React.ReactNode[] & React.ReactNode);
+export type ReactChildren = React.ReactNode | (React.ReactNode[] & React.ReactNode);
```
maheshkulkarni4, line 11:
```suggestion
export type ReactChildren = React.ReactNode[] & React.ReactNode;
```
Doesn't this cover all?

src/datagrid/DataGridFilter.tsx
Bhatii: Sure, I can revert this. but reverting this to previous format like 
`@param {<variable>} <description>` is completely wrong as var name can not take the place of type.
The format suggested by JSDoc is something like `@param {type} <variable>- <description>` 
Please see https://jsdoc.app/tags-param.html and suggest what could be the best course of action.

src/datagrid/DataGridUtils.ts
Bhatii: It is implemented previously.
In this PR there are changes related to position only, so previous implementations are covering these new scenarios in term of UI.

src/datagrid/DataGridFilter.tsx
akate1: Please follow existing implemented pattern

src/datagrid/CustomFilterMulti.tsx
akate1: @Bhatii Please update this code with return statement

src/datagrid/DataGridConstants.ts
```diff
@@ -0,0 +1,19 @@
...
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+/**
+ * Constant for Icon center position offset in pixels
```
akate1, line 12: if this icon height only for data grid filter then please move it to respective file

akate1 (approved):

PR #255, opened 2022-10-28
[OBSDEF-19780] Datagrid footer

Malayvarma27:
This PR is for the ticket [OBSDEF-19780] -> upgrade -> pt#3 (in description)

Issue: 
Datagrid footer item text is not getting updated even after providing correct props.

Resolution: 
Footer item text is now taken from props itself. Earlier it was taken from state. Now removed state for item text.

Tested by patching it to the local workspace. Working fine

Storybook link: http://10.249.253.4:6006

### DIFF
--- a/src/datagrid/DataGrid.tsx
+++ b/src/datagrid/DataGrid.tsx
@@ -289,7 +289,6 @@ type DataGridState = {
     selectAll: boolean;
     allColumns: DataGridColumn[];
     allRows: DataGridRow[];
-    itemText: string;
     pagination?: DataGridPaginationState;
     isLoading: boolean;
 };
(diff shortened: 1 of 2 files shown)


src/datagrid/DataGrid.tsx
```diff
@@ -289,7 +289,6 @@ type DataGridState = {
     selectAll: boolean;
     allColumns: DataGridColumn[];
     allRows: DataGridRow[];
-    itemText: string;
```
akate1, line 292: Please remove documentation as well

akate1 (approved):
Tested
LGTM

src/datagrid/DataGrid.tsx
Malayvarma27: done

m-manaswini (approved):

Issue #252, opened 2022-09-11
 Property 'title' is missing in type '{ children: string; }' but required in type 'Readonly<MainContainerProps>'


PushpendraKatoria:
Getting this error while using MainContainer in App.tsx

TS2769: No overload matches this call.
  Overload 1 of 2, '(props: MainContainerProps | Readonly<MainContainerProps>): MainContainer', gave the following error.
    Property 'title' is missing in type '{ children: string; }' but required in type 'Readonly<MainContainerProps>'.
  Overload 2 of 2, '(props: MainContainerProps, context: any): MainContainer', gave the following error.
    Property 'title' is missing in type '{ children: string; }' but required in type 'Readonly<MainContainerProps>'.
    20 |     render() {
    21 |         return(
  > 22 |             <MainContainer>
       |              ^^^^^^^^^^^^^
    23 |                 Hello
    24 |             </MainContainer>
    25 |         );

squidjam:
This is also happening to me.  Following the documentation leads to an error.

PR #261, opened 2023-01-17
Fix MainContainer issue

squidjam:
Make title optional as it's used in documentation
Make Header component render dependent of title being available Add children to props as JSX.Element or string to support content correctly Code passed through prettier

Signed-off-by: Rebecca Frances <rebecca.francescansinos@materialise.com>


squidjam:
Solves issue [#252](https://github.com/EMCECS/clarity-react/issues/252)

PR #266, opened 2023-04-17
[OBSDEF-25568] Datagrid Filter overlay fix

Malayvarma27:
In this PR we have fixed the filter and it is now appearing on the top of the container and Footer. Also we have implemented that when ever we scroll on the page the filter gets close.

image

Test Link:
http://10.249.253.4:3003/


Bhatii (approved):
Verified on the link, The filter is coming over the datagrid and on scroll it is disappearing. Found working as expected
image
image
image

ShitalSD (approved):
LGTM, verified date filter on Accounts, Groups, Management users tabs

PR #268, opened 2023-04-27
[OBSDEF-25566] DatePicker - Added disabled prop and componentDidUpdate method

Malayvarma27:
Added disable prop to disable datepicker input and calendar button.
Added componentDidUpdate method to update value of input when defaultValue is passes.

storybook : http://10.249.253.4:6006

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@dellstorage/clarity-react",
-    "version": "1.2.4",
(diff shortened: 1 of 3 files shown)


tejasjc-ecs (approved):
@Malayvarma27 please increase the version of clarity-react

psivakumaran (approved):

PR #269, opened 2024-01-16
[OBSDEF-38858] Adding values for datagrid filter

GauravMenezes-Professional:
### Fixes:
- Bumped version of clarity-react
- Added check on scroll event for Datagrid filter. It will now only close if the scroll event is not caused by input

### RCA:

There is a problem currently where adding a long input which does not fit the datagrid filter box makes a scroll event.

Because of this, inputs that are large cannot be added to DatagridFilter.

However, scroll listener was added to close the DatagridFilter on scrolling up or down as without it, the datagrid filter was detaching from the datagrid itself.

Hence modified the code to not close filter if input exceeds datagrid size

### JIRA
https://jira.cec.lab.emc.com/browse/OBSDEF-38858

### Testing:

Jumpbox Link: http://10.227.242.14:3000
Credentials: root/ChangeMe@123 

### Before fix:

OBSDEF-38858

### After fix:

OBSDEF-38858-fixed

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@dellstorage/clarity-react",
-    "version": "1.2.5",
+    "version": "1.2.7",
     "description": "React components for Clarity UI",
     "license": "Apache-2.0",
     "private": false,
--- a/src/datagrid/DataGridFilter.tsx
+++ b/src/datagrid/DataGridFilter.tsx
@@ -234,10 +234,13 @@ export class DataGridFilter extends React.PureComponent<DataGridFilterProps, Dat
         this.toggle();
     };
 
-    private closeFilter = () => {
-        this.setState({
-            isOpen: false,
-        });
+    private closeFilter = (evt: any) => {
+        // This should make sure that only scroll events not due to large inputs are handled
+        if (evt.target.nodeName !== "INPUT") {
+            this.setState({
+                isOpen: false,
+            });
+        }
     };
 
     private toggle() {
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,2 +1,2 @@
 # CODEOWNERS file
-* @akate1 @maheshkulkarni4 @girishdudhe @spiegela @ben-schumacher
+* @akate1 @maheshkulkarni4 @girishdudhe @spiegela @ben-schumacher @ShitalSD @psivakumaran @m-manaswini @Bhatii


m-manaswini (approved):
LGTM

package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@dellstorage/clarity-react",
-    "version": "1.2.5",
```
maheshkulkarni4, line 3: Why did we jump to 1.2.7?  Why not 1.2.6.
So local Bumping pushed it ?

src/datagrid/DataGridFilter.tsx
```diff
@@ -234,10 +234,13 @@ export class DataGridFilter extends React.PureComponent<DataGridFilterProps, Dat
         this.toggle();
     };
 
-    private closeFilter = () => {
-        this.setState({
-            isOpen: false,
-        });
+    private closeFilter = (evt: any) => {
```
maheshkulkarni4, line 237: When the user clicks outside the element, closing the filter is an expected behaviour.  What are we fixing here ?
Is there any side effect or change in behaviour of this on Input of other elements?

maheshkulkarni4:
Please add a GIF showing the Before and After effects of your change.

maheshkulkarni4 (approved):
The change looks very small, but please check my comments and try to answer/address the concerns before merging.

src/datagrid/DataGridFilter.tsx
GauravMenezes-Professional: There is a problem currently where adding a long input which does not fit the datagrid filter box makes a scroll event.

Because of this, inputs that are large cannot be added to DatagridFilter.

However, scroll listener was added to close the DatagridFilter on scrolling up or down as without it, the datagrid filter was detaching from the datagrid itself.

Hence modified the code to not close filter if input exceeds datagrid size

package.json
GauravMenezes-Professional: Checking the package.json in ObjectScale Portal, I see that version attached is 1.2.6, but version in main is 1.2.5. Hence updated version to +1 as I believe one version got skipped.

Is it safe to keep as 1.2.6 instead?

src/datagrid/DataGridFilter.tsx
maheshkulkarni4: Thanks for summarizing.  This is a good RCA to be added on JIRA and PR too.

package.json
maheshkulkarni4: I remember the issue with publishing now.  So normally when we run the git job to publish to npm, it will bump the version as well, but it's not committing back the version on our change.  

So even if the repo shows 1.2.5, the actual published version is 1.2.6
In your case, the repo shows 1.2.7 but, when you publish, it will publish 1.2.8

Just after publishing, please raise another PR to bump the version to 1.2.8 for future.

